Issue #5, opened 2015-09-16
Change the name of the pdf


antoine-lizee:
In the first published article, the name of the pdf has not been changed, which makes the downloaded file have a less than optimal name ("article"). I'm wondering if changing the name to "your_article_name.md" would be suggestive enough for the user of the submission repo to actually change the name of the file!

antoine-lizee:
The proposed PR changes the name of the files to something more obviously changeable. The solution could be totally different though, like enforcing to change the file name during the review process, or even better, a baked-in change of the downloadable file name when linking from the website to the pdf of the repo (I have no idea how to do that though).

rougier:
Thanks. PR merged.

PR #6, opened 2015-09-16
change the name of the article files

antoine-lizee:
Proposition related to #5


rougier:
Thanks. We might need to find some naming scheme for future submission.
Maybe something as "ReScience-Author1-Author2-...-AuthorN-Year"

PR #23, opened 2016-10-16
Added a PR template to ease submission

rougier:
Comments welcome (do we need more information ?)


eddelbuettel:
LGTM -- but shouldn't we also change README.md to reference it?

rougier:
Oops, yes, you're right !

eddelbuettel:
LGTM. It quotes it in almost full in `README.md`; I was thinking more of a file link. But added clarity never hurt.

Ready to merge; not sure if you want a second on it.

rougier:
Thanks @eddelbuettel. I think one review is enough.

PR #27, opened 2017-02-22
Review Request: Stollmeier

Fjanks:
**AUTHOR**

Dear @ReScience/editors,

### Original article

**Title:**  A simple rule for the evolution of cooperation on graphs and social networks
**Author(s):**  H. Ohtsuki, C. Hauert, E. Lieberman and M.A. Nowak
**Journal (or Conference):**  Nature
**Year:**  2006
**DOI:**  10.1038/nature04605
**PDF:**   http://projects.iq.harvard.edu/files/ped/files/nature06a_0.pdf (SI: http://www.nature.com/nature/journal/v441/n7092/extref/nature04605-s1.pdf)

### Replication

**Author(s)**:   F. Stollmeier (ORCID: 0000-0003-4858-0895)
**Repository**:  https://github.com/Fjanks/ReScience-submission/tree/STOLLMEIER-2017
**PDF**:  https://github.com/Fjanks/ReScience-submission/blob/STOLLMEIER-2017/article/stollmeier-2017.pdf
**Keywords**:  Evolutionary Game Theory, Networks
**Language**:   English
**Domain**:  

### Results

---

**EDITOR**

* [x] Editor acknowledgement (@ctb) `February 24, 2017`
* ~~Reviewer 1 (@AdamRTomkins) `February 27, 2017`~~
* [x] Reviewer 2 (@anyaevostinar) `February 28, 2017`
* [x] Reviewer 3 (@KamilSJaron) `June 15, 2017`
* [x] Review 1 decision [**accept**] `April 19, 2017`
* [x] Review 2 decision [**accept**] `July 1, 2017`
* [x] Editor decision [**accept**] `July 4, 2017`


rougier:
Thanks for your submission, we'll assign an editor soon and will proceed with review.
Don't hesitate to ask for update.

rougier:
@Fjanks By the way, do you have a link to a free PDF of the original article ?

rougier:
@ctb Could you edit this submission ?

rougier:
@Fjanks And I forgot, can you add your ORCID id ? (you should be able to update your first post)

Fjanks:
@rougier
I replaced the link to the original article with a link to a free pdf and also added a link to the supplementary. 
I put the ORCID id after my name in the first post. Is that ok or is there somewhere a field for the ORCID id which I didn't see?

ctb:
@rougier I can edit!

AdamRTomkins:
I would happily act as a reviewer for this submission.

rougier:
@Fjanks perfect !

rougier:
@AdamRTomkins thanks !

ctb:
@AdamRTomkins please go ahead! I'm looking for another reviewer now, but there's no reason not to start :).

For your reference, here are reviewer guidelines: https://github.com/ReScience/ReScience/blob/master/reviewer-guidelines.md

ctb:
@anyaevostinar has also agreed to review! @rougier do you need to do anything to give her perms? her ORCID is orcid.org/0000-0001-7216-5283 - thanks, Anya!

rougier:
@ctb @anyaevostinar Simplest would be to register as reviewer [here](https://github.com/ReScience/ReScience/issues/27)

@ctb Don't forget to update PR labels and information at the start of the htread (reviewer names and date)

anyaevostinar:
@ctb @rougier registered

anyaevostinar:
I'm having some problems getting "Preparations" to run. I assumed I should enter: `jupyter notebook` on the command line and then click `replication.ipynb` and then start executing. I've installed all the packages it looks like I need, but am getting:

```
---------------------------------------------------------------------------
DistutilsExecError                        Traceback (most recent call last)
/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/unixccompiler.py in _compile(self, obj, src, ext, cc_args, extra_postargs, pp_opts)
    115             self.spawn(compiler_so + cc_args + [src, '-o', obj] +
--> 116                        extra_postargs)
    117         except DistutilsExecError as msg:

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/ccompiler.py in spawn(self, cmd)
    908     def spawn(self, cmd):
--> 909         spawn(cmd, dry_run=self.dry_run)
    910 

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/spawn.py in spawn(cmd, search_path, verbose, dry_run)
     35     if os.name == 'posix':
---> 36         _spawn_posix(cmd, search_path, dry_run=dry_run)
     37     elif os.name == 'nt':

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/spawn.py in _spawn_posix(cmd, search_path, verbose, dry_run)
    161                           "command %r failed with exit status %d"
--> 162                           % (cmd, exit_status))
    163             elif os.WIFSTOPPED(status):

DistutilsExecError: command 'gcc-4.2' failed with exit status 1
```

Fjanks:
That looks like the compilation of the cython code fails. Did you check whether cython in general works? You could test it with the examples on http://docs.cython.org/en/latest/src/quickstart/build.html.

By the way, I saw in the error message that you use python 3. That reminded me that I totally forgot to make the code compatible with python 3, which I did some minutes ago.

anyaevostinar:
@Fjanks I have gone through the hello world example and the jupyter example in the cython docs and that all works as expected. I've tried again with Python3 and am getting the same:

```
---------------------------------------------------------------------------
DistutilsExecError                        Traceback (most recent call last)
/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/unixccompiler.py in _compile(self, obj, src, ext, cc_args, extra_postargs, pp_opts)
    115             self.spawn(compiler_so + cc_args + [src, '-o', obj] +
--> 116                        extra_postargs)
    117         except DistutilsExecError as msg:

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/ccompiler.py in spawn(self, cmd)
    908     def spawn(self, cmd):
--> 909         spawn(cmd, dry_run=self.dry_run)
    910 

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/spawn.py in spawn(cmd, search_path, verbose, dry_run)
     35     if os.name == 'posix':
---> 36         _spawn_posix(cmd, search_path, dry_run=dry_run)
     37     elif os.name == 'nt':

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/spawn.py in _spawn_posix(cmd, search_path, verbose, dry_run)
    161                           "command %r failed with exit status %d"
--> 162                           % (cmd, exit_status))
    163             elif os.WIFSTOPPED(status):

DistutilsExecError: command 'gcc-4.2' failed with exit status 1

During handling of the above exception, another exception occurred:

CompileError                              Traceback (most recent call last)
/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/pyximport/pyximport.py in load_module(name, pyxfilename, pyxbuild_dir, is_package, build_inplace, language_level, so_path)
    215             so_path = build_module(module_name, pyxfilename, pyxbuild_dir,
--> 216                                    inplace=build_inplace, language_level=language_level)
    217         mod = imp.load_dynamic(name, so_path)

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/pyximport/pyximport.py in build_module(name, pyxfilename, pyxbuild_dir, inplace, language_level)
    191                                   inplace=inplace,
--> 192                                   reload_support=pyxargs.reload_support)
    193     assert os.path.exists(so_path), "Cannot find: %s" % so_path

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/pyximport/pyxbuild.py in pyx_to_dll(filename, ext, force_rebuild, build_in_temp, pyxbuild_dir, setup_args, reload_support, inplace)
    101         obj_build_ext = dist.get_command_obj("build_ext")
--> 102         dist.run_commands()
    103         so_path = obj_build_ext.get_outputs()[0]

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/dist.py in run_commands(self)
    954         for cmd in self.commands:
--> 955             self.run_command(cmd)
    956 

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/dist.py in run_command(self, command)
    973         cmd_obj.ensure_finalized()
--> 974         cmd_obj.run()
    975         self.have_run[command] = 1

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/Cython/Distutils/old_build_ext.py in run(self)
    184 
--> 185         _build_ext.build_ext.run(self)
    186 

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/command/build_ext.py in run(self)
    338         # Now actually compile and link everything.
--> 339         self.build_extensions()
    340 

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/Cython/Distutils/old_build_ext.py in build_extensions(self)
    192             ext.sources = self.cython_sources(ext.sources, ext)
--> 193             self.build_extension(ext)
    194 

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/command/build_ext.py in build_extension(self, ext)
    502                                          extra_postargs=extra_args,
--> 503                                          depends=ext.depends)
    504 

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/ccompiler.py in compile(self, sources, output_dir, macros, include_dirs, debug, extra_preargs, extra_postargs, depends)
    573                 continue
--> 574             self._compile(obj, src, ext, cc_args, extra_postargs, pp_opts)
    575 

/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/unixccompiler.py in _compile(self, obj, src, ext, cc_args, extra_postargs, pp_opts)
    117         except DistutilsExecError as msg:
--> 118             raise CompileError(msg)
    119 

CompileError: command 'gcc-4.2' failed with exit status 1
```

rougier:
@anyaevostinar My 2 cents but do you have `gcc-4.2` installed ?

Fjanks:
If the examples in the cython docs work then I guess that pyximport
is the problem. To test this you could use the hello.pyx from the examples and try to import it in python with 
> import pyximport; pyximport.install()
> import hello  

If that fails maybe it helps to reinstall pyximport. As an alternative, you can do it completely without pyximport, the algorithms.pyx can also be compiled with distutils in
the same way as described for the hello.pyx in the docs. Just replace the
filename 'hello.pyx' by 'algorithms.pyx' in the setup.py script. If that works
remove the line "import pyximport; pyximport.install()" in the
jupyter notebook.

anyaevostinar:
@rougier Yes, though here is the full version info:
`Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 7.0.2 (clang-700.1.81)
Target: x86_64-apple-darwin14.5.0
Thread model: posix`

@Fjanks That pyximport code works correctly with only 'unused function' warnings. I'll try without pyximport anyway.

anyaevostinar:
When building algorithms with distutils I get:
`algorithms.c:232:10: fatal error: 'numpy/arrayobject.h' file not found
#include "numpy/arrayobject.h"`

I can import numpy into Python fine, so I'm not sure what the problem is.

Fjanks:
Ok, thats not what I expected, but this error message is much more informative than the one before.
It doesn't find the numpy header files automatically. The good news is that numpy itself knows where they are, it has a function numpy.get_include(). 
Please try this setup.py:

> from distutils.core import setup
> from Cython.Build import cythonize
> import numpy
> 
> setup(
>   name = 'algorithms',
>   ext_modules = cythonize("algorithms.pyx", include_path = [numpy.get_include()]), 
> )

anyaevostinar:
That does not seem to fix the error, when I try with python3 I get this:

`python3 setup.py build_ext --inplace
running build_ext
building 'algorithms' extension
creating build/temp.macosx-10.6-intel-3.4
gcc-4.2 -fno-strict-aliasing -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -arch i386 -arch x86_64 -g -I/Library/Frameworks/Python.framework/Versions/3.4/include/python3.4m -c algorithms.c -o build/temp.macosx-10.6-intel-3.4/algorithms.o
algorithms.c:232:31: error: numpy/arrayobject.h: No such file or directory
algorithms.c:233:31: error: numpy/ufuncobject.h: No such file or directory
algorithms.c:544: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘__pyx_t_5numpy_int8_t’
`

I copied exactly the setup.py code you specify and placed it in the setup.py file in the code folder and then ran python3 setup.py build_ext --inplace as specified in the tutorial you linked earlier.

Fjanks:
Do you use OS X? I've read that distutils on OS X ignores the include_path. There are workarounds, but since I have only linux systems I cannot test them. Lets try it without distutils.

First, run cython to create the C code:
```bash
cython algorithms.pyx
 ```
Second, start python3 and get the include-path of numpy:
```python
import numpy
numpy.get_include() 
```
Third, compile it with the following command (replace /path/to/numpy with the result from numpy.get_include()):
```bash
gcc-4.2 -shared -pthread -fPIC -fwrapv -O2 -Wall -fno-strict-aliasing -I/Library/Frameworks/Python.framework/Versions/3.4/include/python3.4m -I/path/to/numpy -o algorithms.so algorithms.c 
```

anyaevostinar:
I am using OS X. When running this code, I get:

`$ gcc-4.2 -shared -pthread -fPIC -fwrapv -O2 -Wall -fno-strict-aliasing -I/Library/Frameworks/Python.framework/Versions/3.4/include/python3.4m -I/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/numpy/core/include -o algorithms.so algorithms.c 
couldn't understand kern.osversion `14.5.0'
In file included from /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/numpy/core/include/numpy/ndarraytypes.h:1788,
                 from /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/numpy/core/include/numpy/ndarrayobject.h:18,
                 from /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/numpy/core/include/numpy/arrayobject.h:4,
                 from algorithms.c:435:
/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:15:2: warning: #warning "Using deprecated NumPy API, disable it by " "#defining NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION"
algorithms.c: In function ‘__Pyx_WriteUnraisable’:
algorithms.c:25566: warning: ‘state’ may be used uninitialized in this function
ld: warning: object file (/var/folders/md/pmh2w9w924g1z0nsd4fqr7th0000gn/T//ccrJp9W5.o) was built for newer OSX version (10.10) than being linked (10.4)
Undefined symbols for architecture x86_64:
  "_PyBaseObject_Type", referenced from:
...
ld: symbol(s) not found for architecture x86_64
collect2: ld returned 1 exit status`

With a lot listed in the '...'

rougier:
You can try `python3-config --cflags` and `python3-config --ldflags` for compiler flags:

```
 gcc-4.2 -shared `python3-config --cflags` `python3-config --ldflags` \
         -I`python -c "import numpy; print(numpy.get_include())"` \
         -o algorithms.so algorithms.c
```

anyaevostinar:
@rougier That worked, running the notebook now. Thanks!

anyaevostinar:
The graphs from saved data are all working but when I try to run the simulation with reduced n_graphs and n_runs, I get: "Widget Javascript not detected.  It may not be installed or enabled properly."

Fjanks:
I'm glad it works now!
@rougier: Thank you for the suggestion!
@anyaevostinar: The only javascript widget in the notebook is the progress bar. It is nice to see how long the simulation will take, but the simulations should also run without it. You can either ignore the message or install ipywidgets and activate it:
```bash
pip install ipywidgets
jupyter nbextension enable --py --sys-prefix widgetsnbextension
 ```

**Comment by ctb** [comment]
Hello everyone, how is the review going? :)

**Comment by anyaevostinar** [comment]
@ctb My decision is accept with modification. 

The original paper is one with a valuable contribution without open source code and therefore this replication with provided code is valuable to the scientific community. However, as documented in this thread, there were many problems getting the code to run on OS X, which is a platform used by many scientists who aren't as experienced with computers. In order for this replication to be most useful, a more extensive README is needed that details exact commands and the trouble shooting above. With that addition, I recommend this replication paper to be accepted.

**Comment by rougier** [comment]
@ctb @Fjanks Any update on this submission?

**Comment by Fjanks** [comment]
I'm going to add information for running the code on OS X, but I thought its better to wait for the other referee's comments before I change something.

**Comment by ctb** [comment]
Hi @AdamRTomkins I think we're waiting for your review - any chance you can get it in this week? Please also see https://github.com/ReScience/ReScience-submission/pull/27#issuecomment-295288115.

**Comment by ctb** [comment]
@AdamRTomkins ping?

**Comment by rougier** [comment]
@ctb Maybe we need to pick another reviewer.

**Comment by ctb** [comment]
Hi @Fjanks if you fix the issues that @anyaevostinar I will do a quick review as well. @rougier do we need two formal reviews or will 1 + editor suffice?

**Comment by rougier** [comment]
Two would be better. You can to call for a new reviewer by using the tag (@) ReScience/reviewers
(By the way, don't forget to answer https://github.com/ReScience/ReScience-article-2/issues/1)

**Comment by Fjanks** [comment]
I added the instruction for the manual compilation to the README. 
@anyaevostinar: Is that sufficient for the compilation problem? Is there anything else I can add to help users run the code?

**Comment by ctb** [comment]
@Fjanks thanks! Will ask for more reviewers next comment.

**Comment by ctb** [comment]
@ReScience/reviewers can anyone do a review of this article, please? Thanks!

**Comment by KamilSJaron** [comment]
I can review.
--- edit ---
I can review tomorrow.

**Comment by pietromarchesi** [comment]
I can review!

**Comment by ctb** [comment]
@pietromarchesi @KamilSJaron if one or both of you could give a look that would be excellent! thank you!

**Comment by pietromarchesi** [comment]
I can have a look now, but I wouldn't want to step on @KamilSJaron 's feet, as he was the first to reply.

**Comment by KamilSJaron** [comment]
@pietromarchesi no worries. If you can do the review faster, just go for it.

I have some issues with cython. I installed python/cython though conda. But when I try to compile the .c file generated from the .pyx using `cythonize("hello.pyx")`, I get
```
ld: library not found for -lpython3.5m
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
it is completely unrelated to the manuscrupt, since I am getting the same error for hello world example.

**Comment by dimpase** [comment]
the building procedure looks incredibly iffy to me. IMHO a proper way to build Cython projects is via a setup.py, not by doing something in a notebook cell, and then, if it fails, do some manual compilation, with an instruction that does not take into account the build tools of the reproducer. E.g. in the last comment it looks as if @KamilSJaron has a clang-based toolchain; surely it won't mix well with the gcc, as suggested in the submission.

**Comment by KamilSJaron** [comment]
All right, the original problem was that conda was trying to use gcc from Xcode (which is just a linked clang, therefore the error message). When I installed gcc through conda to environment, it worked. Simulation is running now.

An easy and petty safe way how to reproduce sim on OS X is:
1. install python through conda
2. create an enviroment
3. install cython, gcc, jupyter, (again using conda)...
4. compile cython code (`setup.py` would be indeed comfortable, but I do not find it crucial. The problem I had would not be solved though setup.)
5. run notebook

this should work regardless of what is already installed on computer. Also, note that this is not problem of the paper, but problem of OS X compiler being problematic with cython.

**Comment by dimpase** [comment]
Apple's clang, or other clang versions work just fine with cython. E.g. in
sagemath project we are able to build a hundred or so cython extensions
using clang on OSX, or on Linux, or on FreeBSD.

But it does require proper configuration, not "try stuff in jupyter
notebook, and if it fails, execute this convoluted shell command that would
hopefully work" kind of approach we see here. The latter is a sure road to
irreproducibility, totally not future-proof one.

I think any nontrivial to install python code must be delivered in a form
of proper source Python package, that can be installed using pip or by
running python on setup.py.

**Comment by ctb** [comment]
Thanks - this is really helpful feedback!

**Comment by KamilSJaron** [comment]
So far, I have **not** been able to re-simulate the basic results shown in the manuscript (the analysis of cost), when I run commands in the jupyter notebook. Instead of nice straight lines I get this kind of spaghetti plot (I run it with original setting - `N = 500, k = 10, w = 0.01, n_graphs = 1000, n_runs = 1000`) :
cost_value_estimation

Now I am rerunning the sim to find out if I can replicate this unexpected behaviour. Do you have any idea @Fjanks what might gone wrong?

**Comment by Fjanks** [comment]
@KamilSJaron: I would expect a plot like that with very small values for n_graphs and n_runs. I just tried n_graphs = 10 and n_runs = 100 and the result looks similar. 
Please check these parameters after the second run finished.

**Comment by Fjanks** [comment]
Regarding the compilation:
I don't mind using setup.py instead of pyximport. The reason for pyximport was just that the user could run the code immedeately. When anyaevostinar said it does not work on OS X we tried a setup.py. But as you can see above it caused the same problem, which is why we ended up with this unusual compilation. 
Since I have only linux I would need your help with this. What I found out so far is that on OS X the include_dirs argument in setup(...) is ignored and should instead be passed to Extension(...) (see this [issue](https://github.com/hmmlearn/hmmlearn/issues/43)). So I assume the following setup.py would do it:

```python
from distutils.core import setup
from distutils.extension import Extension
from Cython.Build import cythonize
import numpy

extensions = [Extension("algorithms", ["algorithms.pyx"], include_dirs=[numpy.get_include()])]

setup( ext_modules = cythonize(extensions) )
```
Can someone confirm that this works on OS X?

KamilSJaron:
@Fjanks I tested your setup and it works.

I need to check again how I run the sim.

KamilSJaron:
@Fjanks The graph after rerunning still looks moreless the same and now I am sure that I executed exactly what was in the jupyter notebook without any modification (both `n_graphs` and `n_runs` were set to 1000 and computing time was on a single node more than 8 hours). Can it be something Mac-related?

cost_value_estimation

In the meantime I will try to simulate some primitive cases (like graphs of two nodes, the prob should always converge to 0.5, right?).

code/algorithms.pyx
```diff
@@ -0,0 +1,380 @@
...
+    n_runs: The number of simulations on each graph.
+    benefit_to_cost_ratios: List of b/c ratios.
+    dview (optional): ipyparallel.client.view.DirectView object for parallelization. The default value is None, meaning no parallelization.
+    iteration_type (optional): Either 1 (death-birth process) or 0 (imitation process). Default is 1. 
+    
+    Returns:
+    Fixation probabilities as a numpy-array of the same length as benefit_to_cost_ratios.
+    '''
+    fixation_probabilities = []
+    seeds = [rand() for i in xrange(n_graphs)]
```
KamilSJaron: here you create a set of seeds that is then used for graphs of each bc ratio, which means that bc ratio measurements are actually not independent. I do not think it will change a lot, but if the simulation have used random number for the same things in same order it might change something.
You can just move this line inside of the for loop.

code/algorithms.pyx
```diff
@@ -0,0 +1,380 @@
...
+    The ratio of the number of simulations that ended with a fixation of cooperators.
+    '''
+    srand(seed)
+    cdef long int [:,:] graph = graph_nd
+    cdef int N = graph.shape[0]
+    cdef int first_cooperator_index
+    cdef int C_fixations = 0
+    cdef int i = 0
+    for i in range(n_runs):
+        first_cooperator_index = int((rand() / (float(INT_MAX) + 1) * N))
```
KamilSJaron: I do not expect this to be a problem for the paper, but I thought it is worth of a comment that `rand()` is depreciated. There are other (more elegant) ways how to generate random numbers in C++ (I am not entirely sure how it would work in cython).

I found extremly useful this lecture
https://www.youtube.com/watch?v=LDPMpc-ENqY

Using [Mersenne Twister](https://en.wikipedia.org/wiki/Mersenne_Twister) algorithm for random numbers would also ensure seed-based reproducibility on different platforms / compilers.

Fjanks:
Thanks for pointing at that part of the code! I think you found the problem. 
If I use always the same seed, e.g. seeds = [1 for i in xrange(n_graphs)], I get results that look very much like yours.
I have to read a bit to understand how to generate random number more reliably.

KamilSJaron:
Just to add. The code looks really good and it is very easy to tweak it, so I tried to simulate complete graphs of 2, 4 and 8 individuals with no cost and no benefit resulting in expected probabilities of fixation 0.5, 0.25, 0.125. It seems that this basic behaviour is correct.

Now I am going to wait till @Fjanks commit the change of random number generation.

dimpase:
There is a more fundamental randomisation issue in your code: the random
graphs you generate are not uniformly distributed. Basically, you start by
generating a spanning tree for your graph, but the tree you get is biased:
the vertices you add in the beginning will get more edges than ones you add
in the end, meaning that the tree will be, so to speak, closer to being a
"star" than it should be.

One way to get a uniform spanning tree is by doing a random walk on a
complete graph starting from a vertex, and record edges that led to
discovery of new  (for the walker) vertices. Once all the vertices are
discovered, the recorded edges will be a truly uniform random spanning tree.

Fjanks:
@dimpase: Yes, the random graphs generated by create_connected_random_graph(N,k) are biased. I decided to use this algorithm because that is how I understand the description of the random graph generation in the supplementary of the original paper. And if I understood it correctly (their description is not very detailed) only results from this method are comparable to the results in the original paper.

Fjanks:
Ok, I did a bit research on the issue with rand(). As KamilSJaron already mentioned, it is well known that random numbers from rand() have low quality, so I should not have chosen it in the first place. However, a small period, autocorrelation or a non-uniform distribution do not explain KamilSJaron's results, and for the effect I assume to be the root of the problem I did not find a reference. So I made a little test: On my system the first number from rand() varies with different seeds, but on the Mac of a colleague the first number from rand() was the same for different seed values. This effect results in one node of the network being more often chosen as the position of the initial cooperator than the other nodes. Consequently, the average values for the fixation probabilities would require much larger numbers of simulations to approach the expected values.
To avoid these problems I replaced rand() by Mersenne Twister. Also for Mersenne Twister it seems not so trivial to choose a good set of seed values, but it is not platform dependent and I checked that the first few numbers across the seed values are to a good approximation uniformly distributed.  

I also added the setup.py and updated the compilation instructions.

KamilSJaron:
It seems that it was indeed problem of non-random seeding. The simulations look way better now. So far I simulated only few replicates of the fixation probabilities for different cost values and the graph already looks very similar to the one presented in paper (with more noise of course, but it seems that it quickly converge to the presented fixation probabilities).

Concerning random numbers, I will just have a comment. You have seeded the generator with hardcoded seed values for every possible value of `n_graphs`. 
```
seeds = np.linspace(1, 2**32-1, n_graphs, dtype=np.uint32)
```
The good practise is to allow a seed specified in an argument of the simulator's wrapper and if it is not specified you can generate it using time (not optimal either) or non-seedable generators (like `random_device`). However, the current solution is totally fine for reproduction of the results from the original paper.

I will get back once I will re-simulate at least couple of the different graphs.

KamilSJaron:
I managed to reproduce results of @Fjanks (the simulations of 100 individuals). Therefore I believe that @Fjanks successfully reproduced simulations presented in the original paper "A reference implementation of A simple rule for the evolution of cooperation on graphs and social networks.".

@ctb I suggest to accept the manuscript. My objections were incorporated already.

ctb:
Thanks all! it looks like the new decision should be accept based on https://github.com/ReScience/ReScience-submission/pull/27#issuecomment-312428843 - @anyaevostinar any further thoughts? If not, then I will click the magic "accept" button :)

rougier:
@ctb Don't forget to update the post at the top with relevant dates (you'll need them later)

Fjanks:
Today, while working on another project where I use parts of this code, I discovered a bug in the function create_connected_random_graph(). As a consequence, the average degree of the connected random graphs was slightly smaller than expected.
I fixed the bug and updated the figures in the article. The differences in the three affected plots are almost invisible. Just in case someone would run the simulations with large average degrees it could make a significant difference.

ctb:
Accepted! Please lock this PR @rougier as I do not have permissions?

ctb:
https://github.com/ReScience-Archives/Stollmeier-2017

rougier:
I locked it but I think you have permission.

rougier:
@ctb Can you handle the publication?

rougier:
@khinsen Can you handle the publication?

khinsen:
@rougier This article is accepted, so "handling" just means the technicalities of publication, right? In  that case, yes, I can do that.

rougier:
Yes, I think @ctb forgot to do it. It might be easier if we do it now. I'll have more time by next week so I can do it either if your prefer.

khinsen:
**EDITOR**

This submission has been published and will soon appear at http://rescience.github.io/read/

[DOI](https://doi.org/10.5281/zenodo.847732)

PR #28, opened 2017-02-26
Review Request: Rougier

rougier:
**AUTHOR**

Dear @ReScience/editors,

### Original article

**Title:** Weighted Voronoi Stippling 
**Author(s):**  Adrian Secord
**Journal (or Conference):**   International Symposium on Non-photorealistic Animation and
Rendering  
**Year:**  2002  
**DOI:**  10.1145/508530.508537  
**PDF:**   http://mrl.nyu.edu/~ajsecord/npar2002/npar2002_ajsecord_preprint.pdf  

### Replication

**Author(s)**:  Nicolas P. Rougier
**Repository**:  https://github.com/rougier/ReScience-submission/tree/rougier-2017
**PDF**:  https://github.com/rougier/ReScience-submission/blob/rougier-2017/article/Rougier-2017.pdf
**Keywords**:  Python, Stippling, Voronoi, Computer Graphics, Blue Noise
**Language**:   Python
**Domain**:   Signal Processing

### Results

### Potential reviewers

---

**EDITOR**

* [x] Editor acknowledgment (@ThomasA) `February 27, 2017`
* [x] Reviewer 1 (@rth) `February 28, 2017`
* [x] Reviewer 2 (@almarklein) `February 27, 2017`
* [x] Review 1 decision [**accept**] `April 19, 2017`
* [x] Review 2 decision [**accept**] `April 28, 2017`
* [x] Editor decision [**accept**] `May 9, 2017`


rougier:
@khinsen Can you handle this submission (from me) ? I think @ThomasA can edit it but I'll let you decide. I'll now restrict myself to my author role for this submission.

khinsen:
Yes, I'll take over the EIC role for this submission!

@ThomasA can you supervise the reviewing process?

ThomasA:
Yes :+1:
Just a moment while I read up on the procedure as this is my first.

ThomasA:
@rth and @soolijoo can you review this replication?

soolijoo:
Hi, Sorry this is outside my expertise.

ThomasA:
@soolijoo OK, I will try someone else.

ThomasA:
@almarklein can you review this replication?

almarklein:
Sure! Is this ready for review at this point?

rth:
@ThomasA I would be happy to review this.

khinsen:
@almarklein Submissions are supposed to be ready for review, so please go ahead!

ThomasA:
Thanks @rth and @almarklein, you can go ahead with the review.

ThomasA:
Is everything going OK with the review, @rth and @almarklein?

rth:
@ThomasA Sorry about the delay. Will submit a review within the next few days..

almarklein:
Dito; its on my list, I even have the tab open since yesterday.

article/Rougier-2017.md
```diff
@@ -0,0 +1,132 @@
...
+been
+proposed [@AscencioLopez:2010][@Balzer:2009a][@Balzer:2009b][@Bridson:2007]
+that may be more efficient but with lower quality. We thus privileged this
+implementation that is simple and provides high stippling quality.
+
+# Methods
+
+We applied the method proposed in the original paper with some variations due
+to the non-use of the GPU for computing the Voronoi diagram. In the original
+article, author computes the Voronoi diagram by taking advantage of the graphic
```
almarklein: *the* author

article/Rougier-2017.md
```diff
@@ -0,0 +1,132 @@
...
+card (GPU) and the fact that a set of cones seen from above actually represents
+a Voronoi diagram (see [@Hoff:1999]). We use instead
+the [QHull](http://www.qhull.org) library (through
+the [Scipy](https://scipy.github.io) Python package) for computing the Voronoi
```
almarklein: Can you explain the reason for using QHull instead of the GPU approach? Same for why you "... did not use the proposed optimization" below.

article/Rougier-2017.md
```diff
@@ -0,0 +1,132 @@
...
+the
+[original images](http://cs.nyu.edu/~ajsecord/npar2002/StipplingOriginals.zip)
+but did not obtain any response. We thus display here only the output of our
+replication to be compared with the original ones. The climbing shoe (figure
+@fig:shoe) and the corn plant (figure @fig:corn) are very similar to the images
+displayed in the original article. However, for the large and small Peperomia
+plants (figure @fig:large-plant & @fig:small-plant), the output of our
+replication is clearly at a lower quality without having identified the
+cause. Most probably, the limited resolution of the input image may be a
+critical factor and it is not clear if the author used these small resolution
```
almarklein: How likely is it that the cause is related to the differences in the method, e.g. the omission of the optimization mentioned above? (Or was that a purely performance optimization?)

almarklein: At the bottom of this page it seems to be claimed that these are the original images (which I interpret as being the ones that he used): https://cs.nyu.edu/~ajsecord/stipples.html

article/Rougier-2017.md
```diff
@@ -0,0 +1,132 @@
...
+the standard deviation of the area of the Voronoi regions as in the original
+paper since the definition of the rejection criterion was not clear in the
+original article and quite arbitrary.
+
+# Results
+
+We contacted the original author asking for permission to re-use
+the
+[original images](http://cs.nyu.edu/~ajsecord/npar2002/StipplingOriginals.zip)
+but did not obtain any response. We thus display here only the output of our
```
almarklein: I am slightly confused by what you mean here. I suspect that you did use these images to produce the results shown in this paper. Do you mean that you'd like to have re-published the original (color) images in the paper, or that you'd like access to the output images produced by the original algorithm?

code/README.md
```diff
@@ -1,7 +1,56 @@
...
+This replication has been written and tested on OSX 10.12 (Sierra) using the
+following packages:
+
+ * Python 3.6.0
+ * Numpy 1.12.0
+ * Scipy 0.18.1
+ * Matplotlib 2.0.0
+ * tqdm 4.10
+ 
+Original data is not freelty available and you will need to download the
```
almarklein: typo: freelTy

code/run.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
almarklein: I'd prefer a `main.py` or similar, since I cannot run this on Windows :P

code/stippler.py
```diff
@@ -0,0 +1,232 @@
...
+            plt.savefig(pdf_filename)
+            plt.savefig(png_filename)
+
+        if args.display:
+            plt.show()
+
+    # Plot voronoi regions
+    # for region in vor.filtered_regions:
+    #     vertices = vor.vertices[region, :]
+    #     ax.plot(vertices[:, 0], vertices[:, 1], linewidth=.5, color='.5' )
```
almarklein: Commented code; should be removed?

code/stippler.py
```diff
@@ -0,0 +1,232 @@
...
+                plt.savefig(png_filename)
+
+        bar = tqdm.tqdm(total=args.n_iter)
+        animation = FuncAnimation(fig, update,
+                                  repeat=False, frames=args.n_iter-1)
+        plt.show()
+
+    elif not os.path.exists(dat_filename) or args.force:
+        for i in tqdm.trange(args.n_iter):
+            regions, points = stipple(points, density)
```
almarklein: If I run `python stippler.py ..\data\boots.jpg  --save --force --n_point 200 --n_iter 50 --pointsize 0.5 2.5 --figsize 8` the code raises an error here:
```
Traceback (most recent call last):
  File "stippler.py", line 201, in <module>
    regions, points = stipple(points, density)
NameError: name 'stipple' is not defined
```

almarklein: ~~... and trying to put a filename at the end does not work: `stippler.py: error: unrecognized arguments: output.png`~~  oh, my mistake; that positional arg is the input file, nevermind... though the error above is still preventing me from running the code.

rth:
### REVIEWER 1

The proposed reproduction paper is quite interesting and the included code is well structured and very readable. However, two major points should to be addressed, in my option,
  - missing definition of the `stripple` function which prevents executing the python code
  - the difference in the way edges are handles in the resulting images as compared to the original paper should be discussed (cf. detailed comment below).

### Code and installation instructions

* A minor point, it might be preferable upgrading the `tqdm` dependency from `4.10.0` to `4.11.2` as  `4.10.0` does not exist in conda (it exists in conda-forge but only for Python versions <=3.5) .
* Pillow (PIL)  (e.g. `pillow=4.0.0`) is a missing dependency. If not installed one gets the following error,
   ```py.
   Traceback (most recent call last):
     File "stippler.py", line 134, in <module>
       density = 1.0 - scipy.misc.imread(filename, flatten=True, mode='L')/255.0
   AttributeError: module 'scipy.misc' has no attribute 'imread'
   ```
* Minor point, in `run.sh` the path to the image files should probably be `../data/boots.jpg` instead of `data/boots.jpg` etc..
* As already pointed out by @almarklein the `stripple` function is not defined in `stippler.py` and the script cannot be executed.
* It might be useful to indicate the versions of pandoc and texlive used to generate the pdf. The generation of the pdf with the provided  instructions produces the following error, 
  ```
  % biber Rougier-2017     
   [...]                                              
   INFO - Looking for bibtex format file 'bibliography.bib' for section 0
   ERROR - Cannot find 'bibliography.bib'! 
   ```
  since bibliography is written in `Rougier-2017.bib` not `bibliography.bib`
 
The installation instructions were tested on Gentoo Linux with Python installed using conda.

### Article comments 

* It might have been useful to add a sentence discussing the implications (if any) of not using GPU for computing the Voronoi diagram and of not using the optimization for the weighted centroid calculations (i.e. does this only have performance implications, or would it affect the results).
* I would disagree with the explanation given as to why the reproduction of Rougier:Figure 3 & 4 is worse than in the orginal paper
  > Most probably, the
  > limited resolution of the input image may be a critical factor and it is not clear if the
  > author used these small resolution versions or if he used higher resolutions.
  
  A fairly characteristic feature of the original images is a sharp outline of edges, particularly visible in images with a low number of points (Secord:Figure 6), or in Secord:Figure 4. In the original paper there is a discussion of this point,
  > Figure 4 shows a different small Peperomia plant, [..]
  >  In particular, note the hard edges  maintained by the stipple drawing. 
  >  [..] While the method of Deussen et al. can easily produce sharp edges through user in-
  > teraction, producing the gradual change in tone visible on the leaf
  > would be difficult. The even spacing of points along the edges of
  > the leaves is the result of the interaction of the centroidal Voronoi
  > diagram, which attempts to space all points evenly, and the density
  > function ρ, which restricts points to the essentially one-dimensional
  > edge.

  In my opinion, the reproduced images do not exhibit such sharp outline of edges, and additional discussion is necessary to explain this difference. Adding a figure with a low number of points, e.g. to reproduce Secord:Figure 6 might also be useful as this phenomena is particularly visible there.
 * If we look at the detail of the outer right leave in Rougier:Figure2  (left),
   
   deatil_leaf
   
    the points look larger and overlap unlike in the original figure (right) which suggest that stipple dot size might not be exactly the same.

rth:
Also, it would have been extremely helpful if @ajsecord could give permission to use the original stippled images in this reproduction paper, so that they could be visually compared side by side (and with the same resolution).

ajsecord:
Happy to help. What do you need?

rougier:
@almarklein @rth Thank you for the review, I will soon answer and make modification to both the paper and the code.

@ajsecord I sent you a mail asking your for permission to re-use your images to be used in this article and to include them along this manuscript (under a CC0). I did not get an answer (not sure I used the right email address) and  concluded I cannot use them.  Could you consider granting such permission ?

article/Rougier-2017.md
rougier: Corrected.

article/Rougier-2017.md
rougier: GPU approach would require opening a GL context and I thought it would be too much trouble to make the script work on OSX/Linux/Windows. Plus, the GPU suffers from lack of precision because it works at image level (even if authors use higher resolution to cope with such imprecision). Since the qhull library is directly accessible from within scipy, I prefer to use this.

I will add a justification in the text.

As for optimization, I implemented the optimized version that is less readable in my opinion. I'll leave the "old" code for informational purpoose.

article/Rougier-2017.md
rougier: Optimized and non-optimized are supposed to give the exact same results and I checked this is the case.

For the image, they are accessible but you cannot use them without permission.

article/Rougier-2017.md
rougier: Access is ok, but I would have liked to ship the original picture in the data directory of this submission (using CC0) but for this, I need permission from the author. Same for side-by-side comparison of original and replicated results.

code/README.md
rougier: Corrected.

code/run.sh
rougier: Ok, I'll do that.

ajsecord:
@rougier Thanks for your patience—I had missed that email. I just replied with granting of permission.

code/stippler.py
rougier: This serves at showing how you can plot Voronoi cells if you want. I would prefer to keep it as it is.

code/stippler.py
rougier: This is an error from me. I'll double check before pushing modification.

article/Rougier-2017.md
rougier: Get permission in the end. I added the original pictures.

code/run.sh
rougier: Done.

rougier:
I'm now doing the GPU implementation, I will answer all reviews once this is done.

rougier:
@cth @almarklein 

I've made several changes to both the article and the code and hopefully this would address your comments. The biggest change in the code has been to resize the input image to have sufficient resolution (this actually implicitly explained by the author in the article). This is done using the `scipy.ndimage.zoom` function (order 0). This really improved the quality of the output even though some small differences still exist for some images but I'm not sure how this can be fixed.

I've added a discussion on why not using the GPU in the introduction but I'm also adding an experimental GPU implementation for the centroidal voronoi tesselation. It relies on glumpy and I  did not want to add this dependency to the submission. In the end, this implementation would be faster (x3 speedup from my tests) but there are precision problem that would require extra work to be fixed.

Concerning the sharp outline of edges that were previously missing, they are now present in all output even if they might be slightly less clean than in the original.

plant2_400x400-stipple

The difference in leaf shading is quite weak compare to author's version but I don't know the reason. I made some tests with a bigger image of (different) leaves and in this case I think the output is ok, so maybe this is a resolution and/or rescaling problem.

leafs-montage


I also tried to fix the point size to match the originals, but given point sizes in the original articles are not really relevant (no unit) and difficult to interpret when the output is vectorial.

I hope I did not forget to address some comments, if this is the case, just tell me.

boots-montage

rth:
@rougier Thank you for this update! The images do look better. I will look into the changes in the details over the next few days and will get back as soon as I can.

code/README.md
almarklein: not yet, it seems ;)

code/run.py
```diff
@@ -0,0 +1,46 @@
+#! /usr/bin/env python3
+import os
+
+# ---------------------------------------------------------------------------
+# Original images from http://mrl.nyu.edu/~ajsecord/
+# -> http://cs.nyu.edu/~ajsecord/npar2002/StipplingOriginals.zip
+# ---------------------------------------------------------------------------
+
+os.system("./stippler.py ../data/original/plant4h.png --save \
```
almarklein:
```
  File "C:\dev\py\rescience\code\run.py", line 13
    os.system("./stippler.py ../data/original/plant5_700x700.png  --save
                                                                       ^
SyntaxError: EOL while scanning string literal
```

almarklein: sorry, it applies to line 13, where its missing a backslash.

code/run.py
```diff
@@ -0,0 +1,46 @@
...
+           --n_point 20000 --n_iter 50 --pointsize 1.5 1.5 --figsize 6 \
+           --threshold 255 --force --interactive")
+
+os.system("./stippler.py ../data/original/plant5_700x700.png  --save
+           --n_point 20000 --n_iter 50 --pointsize 2.5 2.5 --figsize 8 \
+           --threshold 220 --force --interactive")
+
+os.system("./stippler.py ../data/original/plant2_400x400.png  --save \
```
almarklein: Windows bytes again; forward slash generally works in Windows, but apparently the `./` goes wrong :(   This works:
```py
cmd = '.%sstippler.py' % os.sep

os.system(cmd + " ../data/original/plant4h.png --save \
...
```


```

Comment (almarklein) [reply code/run.py]: Oh, but that also does not work because file associations. Forget the above and do:
```py
import sys
cmd = sys.executable

os.system(cmd + " ./stippler.py ../data/original/plant4h.png --save \
#                /\  mind the space
...
```

article/Rougier-2017.md
```diff
@@ -0,0 +1,160 @@
...
+We display only the output of our replication to be compared with the ones in
+the original articles. The climbing shoe (figure @fig:shoe), the corn plant
+(figure @fig:corn) and the large Peperomia plants (figure @fig:large-plant) are
+very similar to the images displayed in the original article. It it to be noted
+that we used different thresholds in order to obtain white areas similmar to
+the original images. Without such threshold there would be no reason to have
+white areas as for example in the leaves of figure @fig:large-plant.
+
+For the small Peperomia (figure @fig:small-plant), the output of our
+replication may be considered to be not exactly similar. If the hard edges are
```
almarklein: "may be considered to be not exactly similar" sounds a bit overly "careful" :)  perhaps "For the [...] the output of our replication is less similar".

article/Rougier-2017.md
```diff
@@ -0,0 +1,160 @@
...
+maintained by the stipple drawing, the differences in shading in our
+replication is not obvious compared to the original image. Most probably, the
+limited resolution of the input image may be a limiting factor but it is not
+clear if the author used these small resolution versions or if he used higher
+resolutions. We tested this hypothesis using a bigger image showing leaf with
+different shading (figure {#fig:leafs}) and in such a case, shading is visible
+in the output.
```
almarklein: Perhaps @ajsecord has an idea about what caused this discrepancy?

almarklein:
## Revierwer 2

I've left some comments in the code to make `run.py` work on Windows. One other issue is that the process does not move to the next image until the figure that shows the progress is closed (it took me a while to figure this out, I thought that the process had stalled).

Other than that, the results look good. I'm curious to see if the original author has an idea about the source of the remaining inconsistencies.

article/Rougier-2017.md
rougier: Corrected!

code/run.py
rougier: Fixed (hopefully) !

rougier:
@ThomasA @rth @almarklein

I've made corrections by @almarklein, I hope it works better now

ThomasA:
@almarklein can you check if the corrected code executes properly on Windows now?

rth:
### REVIEWER 1

With this latest version, I can confirm that the code runs fine and that the pdf can be generated from sources (on Linux). 

There remains a few minor points,
 * As reported in comments above, in `run.py` ther is a missing backslash line 15,
   ```
      File "code/run.py", line 15
    os.system(cmd + " " + "./stippler.py ../data/original/plant5_700x700.png  --save
                                                                                   ^
    SyntaxError: EOL while scanning string literal
     ```
 * In `README.md` and `code/README.md` the following paragraph can be removed, 
   > Original data is not freely available and you will need to download the
[original images](http://cs.nyu.edu/~ajsecord/npar2002/StipplingOriginals.zip)
prior to running the `run.sh` to obtain results.
  
   since the original images are now included.
 * Missing figure reference in 
   > We tested this hypothesis using a bigger image showing leaf with different shading (figure
{#fig:leafs})

Apart from that I agree that in the original stippled images are now reasonably well reproduced, and I think this paper is ready to be published.

rougier:
Thanks @rth, I've integrated your corrections.

khinsen:
Following a recent discussion (ReScience/ReScience#43) about the non-reproducibility of articles published in ReScience, I did a quick test on this one using the (nearly?) final but not yet published code. As explained in the earlier discussion, all I checked is if I can install the dependencies and run the software - I don't look at the results, since I am not competent in judging them anyway.

Executing `run.py` ends with a crash:

```
Traceback (most recent call last):
  File "./stippler.py", line 138, in <module>
    density = scipy.misc.imread(filename, flatten=True, mode='L')
  File "/Users/hinsen/pyenvs/pydev35/lib/python3.5/site-packages/scipy/misc/pilutil.py", line 154, in imread
    im = Image.open(name)
  File "/Users/hinsen/pyenvs/pydev35/lib/python3.5/site-packages/PIL/Image.py", line 2312, in open
    fp = builtins.open(filename, "rb")
FileNotFoundError: [Errno 2] No such file or directory: '../data/plant.jpg'
```

There is indeed no file `../data/plant.jpg' in the repository.

rth:
> There is indeed no file `../data/plant.jpg' in the repository.

You are right, @khinsen I missed that: stopped the `run.py` before the last image as it was taking a while and since `data/plant.jpg` is not used in the paper, as far as I can tell. Thanks for checking it.

While I understand the limitations of the CI based approach  discussed in the link above (particularly for the long term maintenance), having a Travis CI setup that runs the `run.py` and generates the paper would have allowed to detect this issue,  as well as to make sure that all dependencies are specified, and detecting other issue that came up earlier during review. Even if CI would not be run after publication... 

In this case, it would have also allowed to reduce the overall review delay (by reducing the number of iterations needed between the author and reviewers).

rougier:
Fixed. But how would and where would we setup such a CI ?

khinsen:
@rougier [Travis CI](https://travis-ci.org/) should be able to handle most ReScience papers. There is a time limit of 50 minutes for a job, so in some cases it would not be possible to run a full replication but only a subset.

The main problem I see is the significant additional effort required for setting this up. It might well discourage authors from submitting their work to ReScience if we make it compulsory. But on a voluntary basis, this would be an interesting experiment.

almarklein:
> @almarklein can you check if the corrected code executes properly on Windows now?

I will have a look asap, but it will probably be next week, as I'm a rather busy right now.

ThomasA:
> I will have a look asap, but it will probably be next week, as I'm a rather busy right now.

OK, we will wait. I do not have access to checking it on Windows myself.

article/Rougier-2017.md
```diff
@@ -0,0 +1,159 @@
...
+
+![Leafs (1195x1024), 20 000 dots, 50 iterations, variable point size](./leafs-montage.png){#fig:leafs}
+
+
+# Conclusion
+
+Most of the results have been replicated even though some slight discrepancies
+remain in the final output for one image. Without further contact with the
+original author, it is difficult to identify the precise cause but most likely,
+the problem occurs because of the limited resolution of the input picture. We
```
almarklein: "Without further contact with the original author ..." This raises questions; I'd either ask him or just say "It is difficult to ...".

almarklein:
I can confirm that the code is working properly on Windows. Except for that line in the conclusions that I just commented on, I think that this is ready for publication.

rougier:
@almarklein I've made the correction.

rougier:
@ThomasA Is there anything else needed ?

ThomasA:
Only that I want to read the whole thing through to make sure everything 
is OK :-) Will do today...

khinsen:
@ThomasA Any news on when "today" will be?

ThomasA:
@khinsen sorry, that will be _today_ then. Working on it...

article/Rougier-2017.md
```diff
@@ -0,0 +1,159 @@
...
+# Methods
+
+We applied the method proposed in the original paper with some variations.
+First, we did not use the GPU to generate the Voronoi diagram because this
+would introduce a dependency on OpenGL that would make the script more fragile
+and more difficult to run on all systems (Linux/OSX/Windows). However, we
+provide in the accompanying code an experimental implementation based on
+the [glumpy](http://glumpy.github.io) library. As in the original article, this
+experimental implementation takes advantage of the fact that a set of cones
+seen from above using an orthographic projectionactually represents a Voronoi
```
ThomasA: Can you please correct `projectionactually`?

article/Rougier-2017.md
```diff
@@ -0,0 +1,159 @@
...
+
+$${\bf C}_i = \frac{\int_A {\bf x}\rho({\bf x})dA}{\int_A \rho({\bf x})}$$
+
+Each cell is rasterized (as a set of pixels) and the centroid is computed using
+the optimization proposed by the author that allow to avoid to compute the
+integrals over the whole set of pixels composing the Voronoi cell. As noted by
+the author, the precision of the method is directly related to the size of the
+Voronoi cell. Consequently, if the original density image is too small
+relatively to the number of stipples, there might be quality issues. We used a
+fixed number of iteration ($n=50$) instead of using the difference in the
```
ThomasA: "number of iteration**s**"

article/Rougier-2017.md
```diff
@@ -0,0 +1,159 @@
...
+standard deviation of the area of the Voronoi regions as in the original paper
+since the definition of the rejection criterion was not clear in the original
+article and quite arbitrary.
+
+Last, we added a threshold parameter that allows to perform a pre-processing of
+the densty image. Any pixel with a gray level above the threshold is set to the
```
ThomasA: "the dens**i**ty image"

article/Rougier-2017.md
```diff
@@ -0,0 +1,159 @@
...
+threshold value before normalizing the density image. This was necessary for
+replicating some of the original images (see results section for further
+explanation).
+
+# Results
+
+We display only the output of our replication to be compared with the ones in
+the original articles. The climbing shoe (figure @fig:shoe), the corn plant
+(figure @fig:corn) and the large Peperomia plants (figure @fig:large-plant) are
+very similar to the images displayed in the original article. It it to be noted
```
ThomasA: "it it"?

article/Rougier-2017.md
```diff
@@ -0,0 +1,159 @@
...
+
+![Pot plant (1195x1024), 20 000 dots, 50 iterations, variable point size](./pot-plant-montage.png){#fig:pot-plant}
+
+![Leafs (1195x1024), 20 000 dots, 50 iterations, variable point size](./leafs-montage.png){#fig:leafs}
+
+
+# Conclusion
+
+Most of the results have been replicated even though some slight discrepancies
+remain in the final output for one image that are. It is difficult to identify
```
ThomasA: "that are"?

ThomasA (commented):
A few typo corrections

ThomasA:
@rougier can you please fix the merge conflict? It seems to be caused by `number` introduced in d0c46b0c9fc27bae88e7d1b053eebe149157066f. I guess you should assign number 4 as per https://github.com/ReScience/ReScience/issues/48#issuecomment-300130364 (I hope I am doing it right?)

ThomasA:
@khinsen with a conclusion like 

> We therefore think most of the results have been replicated.

Should this submission be marked as partially replicated?

rougier:
@ThomasA Typos have been corrected. For the merge conflict, it is not really a problem since this PR is not supposed to be merged anyway.

khinsen:
@ThomasA A partial replication is a replication of only some of the results of the original paper. My understanding of the conclusion is different: all results were replicated but for one image the agreement with the original paper is not as good as for the others. Since the reviewers are happy with this, I conclude this is a complete replication within the state of the art of the field.

So unless the author (@rougier) or the reviewers (@rth and @almarklein) disagree with my interpretation, I'd say we can classify this as a full replication.

ThomasA:
@rougier I have trouble rebuilding the article. I have installed pandoc and pandoc-crossref. `make` gives me the following output:
>pandoc-crossref: Error in $: mempty
>pandoc: Error running filter /home/tha/.cabal/bin/pandoc-crossref
>Filter returned error status 1
>Makefile:10: recipe for target 'rescience-template.tex' failed
>make: *** [rescience-template.tex] Error 83

Any idea what is wrong?

ThomasA:
@rougier it seems it was just the newly installed pandoc-crossref that does not play nicely with Ubuntu's supplied version of pandoc.

ThomasA:
@rougier I am stuck trying to build the PDF. I cannot use the 'fontawesome' package on my system. XeLaTeX seems to try to call some font-generating stuff that fails. I have TeXlive 2016 installed in Ubuntu (installed directly from TexLive - not Ubuntu repositories). The fontawesome package **is** installed, but somehow not useable.

ThomasA:
@rougier if I comment out `\usepackage{fontawesome}` from the template, make just seems to stall and never get further than this:

    $ make all
    [1/4] Processing Rougier-2017.tex (pass 1)

rougier:
If you compile using directly the latex command, what is the output ?
The fontawesome package  is needed only for putting the github glyph into the red box on the right.

rougier:
For the fontawesome installation, what is the error in the font generating stuff ?

ThomasA:
@rougier By the way, I just by mistake pushed some updates to your repo. Now I do not understand how I actually had the permission to do so? I guess this is not a major problem - the updates are supposed to go in the archive repo anyway.

ThomasA:
Regarding fonawesome, making the paper says:

    $ make Rougier-2017.pdf
     [1/4] Processing Rougier-2017.tex (pass 1)
    
    kpathsea: Running mktextfm FontAwesome
    /usr/local/texlive/2016/texmf-dist/web2c/mktexnam: Could not map source abbreviation F for FontAwesome.
    /usr/local/texlive/2016/texmf-dist/web2c/mktexnam: Need to update /usr/local/texlive/2016/texmf-dist/fonts/map/fontname/special.map?
    mktextfm: Running mf-nowin -progname=mf \mode:=ljfour; mag:=1; nonstopmode; input FontAwesome
    This is METAFONT, Version 2.7182818 (TeX Live 2016) (preloaded base=mf)
    
    
    kpathsea: Running mktexmf FontAwesome
    ! I can't find file `FontAwesome'.
    <*> ...our; mag:=1; nonstopmode; input FontAwesome
                                                      
    Please type another input file name
    ! Emergency stop.
    <*> ...our; mag:=1; nonstopmode; input FontAwesome
                                                      
    Transcript written on mfput.log.
    grep: FontAwesome.log: Ingen sådan fil eller filkatalog
    mktextfm: `mf-nowin -progname=mf \mode:=ljfour; mag:=1; nonstopmode; input FontAwesome' failed to make FontAwesome.tfm.
    kpathsea: Appending font creation commands to missfont.log.

I am whining about it here as well: https://tex.stackexchange.com/questions/369068/how-can-i-use-the-fontawesome-package-with-xelatex

rougier:
Yes, I can see your changes in my repo. That's totally weird.
In the meantime, once you've made all the changes, I'll be able to build the PDF and push it.

rougier:
Just to be sure, you've also installed the `FontAwesome.otf` somewhere ?

ThomasA:
Yes, it is in the Texlive tree

rougier:
Oh, the font is bundled with the tex package ?
For the push permission, I'll ask github support how this is possible.

ThomasA:
Yes, fontawesome is included in Texlive. I am getting some help here: https://tex.stackexchange.com/questions/369068/how-can-i-use-the-fontawesome-package-with-xelatex, but none of it seems to work...

ThomasA:
OK, I finally made it work.
One remaining oddity is that `xelatex` does not seem to interpret ` `` ''` correctly the way `pdflatex` does. Is that normal?

ThomasA:
I have been looking at this: https://tex.stackexchange.com/questions/193412/what-is-happening-to-the-quotes and this: https://tex.stackexchange.com/questions/278251/ligatures-stopped-working-in-xelatex.
I can fix the quotes if I insert `\defaultfontfeatures{Ligatures=TeX}` after line 33 in 'rescience-template.tex', but only if I run `\setmainfont{[some font]}` afterwards to set a font. Without it it has no effect. If I attempt to insert the `\defaultfontfeatures` line before line 32 in 'rescience-template.tex', make stalls in the first processing of Rougier-2017.tex (doesn't err - just never finishes).
I suspect that `fontawesome` is somehow messing with this `Ligatures` feature...

ThomasA:
Should we simply ignore that ` `` ''` quotes look ugly with this compile workflow?

rougier:
Could you make a PR for the main rescience-template and fix it on this repo then ?
I go the explanation for your commit in my branch, it's a new feature of the PR where you can allow edits from maintainers (there is a tickbox for allowing or not).

ThomasA:
Sorry, I got stuck last week trying to compile the paper for publishing. This week I am swamped with a course I am giving. I hope to pick this back up tomorrow or during the weekend.

rougier:
@ThomasA Any chance to publish it this week ?

khinsen:
@ThomasA a gentle reminder...

ThomasA:
Sorry, I will try to do it tomorrow evening. These days are packed with 
teaching and conducting exams tomorrow...

ThomasA:
@rougier could you please rebuild the PDF on rougier/ReScience-submission on your rougier-2017 branch (including the changes it turned out I was able to push to your repo)? I have no XeTeX experience and seem unable to get the fonts set up correctly on my system to make it compile.
Then I will pull it from there and archive it.

rougier:
Done.

ThomasA:
**EDITOR**

This submission has been published and will soon appear at http://rescience.github.io/read/

[DOI](https://doi.org/10.5281/zenodo.802285)

khinsen:
This looks good! Thanks to @ThomasA for handling the review, to @rth and @almarklein for reviewing, and to @rougier for doing the replication work!

PR #30, opened 2017-04-19
Review Request: Metzner

ChristophMetzner:
Christoph Metzner  

Dear @ReScience/editors,

**AUTHOR**
### Original article

**Title:** Modeling GABA alterations in schizophrenia: a link between impaired inhibition and altered gamma and beta range auditory entrainment
**Author(s):** Vierling-Claassen, D., Siekmeier, P., Stufflebeam, S., & Kopell, N.
**Journal (or Conference):** Journal of Neurophysiology  
**Year:** 2008
**DOI:** 10.1152/jn.00870.2007
**PDF:**  http://jn.physiology.org/content/99/5/2656

### Replication

**Author(s):**   Christoph Metzner
**Repository:**  https://github.com/ChristophMetzner/ReScience-submission/tree/Metzner-2017
**PDF:**  https://github.com/ChristophMetzner/ReScience-submission/tree/Metzner-2017/article/Metzner-2017.pdf
**Keywords:**  Neuroscience, Gamma and Beta Oscillations, Auditory Entrainment, Schizophrenia, Python
**Language:**  Python
**Domain:**  Computational Neuroscience

### Results

### Potential reviewers

---

**EDITOR**

* [x] Editor acknowledgment (@oliviaguest `24 April 2017`)
* [x] Reviewer 1 (@aaronshifman `25 April 2017`)
* [x] Reviewer 2 (@pietromarchesi `25 April 2017`)
* [x] Review 1 decision [**accept**]
* [x] Review 2 decision [**accept**]
* [x] Editor decision [**accept**]


rougier:
@ChristophMetzner Thanks for your submission, we'll assign an editor soon. Do you have a link to an open-access version of the original article ?

rougier:
@otizonaizit Could you handle this submission ?

ChristophMetzner:
@rougier Thank you. Here is a link to the original article http://jn.physiology.org/content/99/5/2656

otizonaizit:
@rougier : Hi, Nicolas! I won't be able to take care of this in a timely manner for the next month or two, so it's probably better to look for a different editor. If you don't find anyone suitable, please come back to me!

rougier:
@otizonaizit Thanks for the quick answer. I'll look for another editor.

aaronshifman:
I might be early: but I'd volunteer as reviewer.

rougier:
@oliviaguest Can you  handle this submission ?

oliviaguest:
Sure.

oliviaguest:
@rougier whose responsibility is it to re-compile the pdf to make sure the internal references show up without question marks and without just outputting the labels, etc.?

rougier:
@oliviaguest The one who asks 😄 More seriously, you can ask author(s) to try to correct for the internal references. Because they know their manuscript much better than anyone else.
Most probably this might be due to trailing whitespaces or a pandoc version mismatch. But if not, we (edtiros) will have to give a closer look.

oliviaguest:
@ChristophMetzner could you take a look at the above issue regarding internal references, please?

oliviaguest:
@pietromarchesi got back to me super quickly — so we have both reviewers ready to go! :+1:

rougier:
@oliviaguest @aaronshifman @pietromarchesi You've been granted a new ⭐️ on the [board page](http://rescience.github.io/board/) (it's a link to the review).

@pietromarchesi Can you give me your ORCID number?

pietromarchesi:
@rougier My ORCID is 0000-0001-5955-6909.

pietromarchesi:
@oliviaguest Can we begin a conversation with the authors and discuss fixes as we encounter problems/imperfections, or should we wait and post a more thorough revision? I have already noticed a few small things that the author could start to fix as I work my way through the rest of the code. 

Also, I cannot seem to clone the [forked repo](https://github.com/ChristophMetzner/ReScience-submission/tree/Metzner-2017). It clones the original ReScience-submission repository, not the forked one (downloading the .zip on the other hand works fine). If anybody knows how to fix this, it would be helpful.

oliviaguest:
@rougier will hopefully know the answers to both your questions @pietromarchesi!

ChristophMetzner:
@oliviaguest I will look at the references asap.

rougier:
@pietromarchesi you have to switch to the author branch and pull again.

pietromarchesi:
(REVIEWER 1) I looked through the submission and noted down some comments.

Preliminaries
-------------

The readme instructs to run `run_example.py`, whereas the file is `example_run.py`.

The method `rasterPlot` of the class `simpleModel` in `simple_model_class.py` takes an additional argument `name` which is not provided on lines 55 and 56 of `example_run.py`.

On line 55 of `run_simulations.py`, the path of the Numpy seed is incorrect: it needs to be changed to `../data/Seeds.npy`. In the line below, the name of the directory has to be changed from '../data_run'
to `../data`.

In `run_simulations.py`, I noticed that the files which contain the average over trials (i.e. `sims_ctrl_avg_20Hz.npy`) which are later called in `plot_figures.py`, are never saved. I added that
 manually on line 84 of `run_simulations.py`:

 ```python
 	filename_ctrl_avg = directory+'/sims_ctrl_avg_'+str(int(drive_frequency))+'Hz'
	filename_schiz_avg = directory+'/sims_schiz_avg_'+str(int(drive_frequency))+'Hz'

	np.save(filename_ctrl_avg, meg_ctrl_avg)
	np.save(filename_schiz_avg, meg_schiz_avg)
 ```

 And then changed the path to the files in `plot_figures.py`, where they are loaded around line 51. I did not make a specific folder for these average-over-trial results, but perhaps it might be cleaner to do so.

 Note that in `plot_figures.py`, the `/data_run` repository has to be renamed to `/data` wherever it appears.

 Finally, I think the file `analysis2.py` is an older and  not commented version of `analysis.py` which should probably be removed.

 I now see that at least some of my previous comments are already corrected in the notebooks, but the changes still need to be transferred to the `.py` files.


Code
----
At the expense of efficiency, the code is very readable, scoring points in article 7 of the zen of Python.

 One minor flaw that I happened to notice is that lines 137 and 150 in `simple_model_class.py` are indented one too many times. This results in the list of spike times being saved every time and not just at the end of the while loop when it's complete.

 Not to be a PEP8 nazi (nobody likes those) but the very long equations in `simpleModel.run` could perhaps be split over two lines, as they are basically double the recommended line length.

Perhaps the comments on lines 167 and 171 in `simple_model_class.py`, which state that the following lines 'seem awfully complicated' could be replaced or integrated with a line that refers to the
 corresponding equation in the article.

Replication
-----------

Overall, the main findings are indeed replicated, however, as recognized by the author, some
of them are slightly less prominent than in the original paper. In this regard it is not clear to me what the author @ChristophMetzner  means when he indicates that this might be due to a difference in noise: where could this difference arise? Are you referring to just stochastic effects or to something else entirely? It might be good to make this a bit clearer in the article as well. 

 If my understanding is correct, the emergence of a 20 Hz component is due to the fact that longer decay of the IPSC's hinders the 40Hz entrainment and you get responses to alternate inputs from the drive. In the absence of noise, your 40Hz-drive schizophrenic network will respond only with 20Hz, and increasing noise brings in the 40Hz component, because some cells still have enough input to fire in spite of the prolonged inhibition. All of this is to say that perhaps scaling the noise a bit lower could give you more of the 20Hz and 40Hz, making the result of the paper more marked. Have you tried anything like that? 


Other
-----
The labels in the plots of the article are a bit too small to be readable.

aaronshifman:
# Code
The code was fairly well commented, however in some places it could do with some more comments. For example generating the spike trains (simple_model_class: line 130-140). 

I'm not sure I understand _getSingleSpikeTimes - what being iterated over? Is that time and state?

I'm also unclear as to what is going on in calculatePSD. If I understand correctly meg is a time series and your slicing starting at an index that is 20% of the sampling frequency? I think this could use more comments

As @ChristophMetzner mentioned he was going for a readable over efficiency model for the code. However there are some places where the explicit nature of the code detracts from readability.

For example (line 128) just scales background_rate by 1000
 `rate_parameter = 1000*(1.0/self.background_rate)`
`rate_parameter = 1.0/rate_parameter`

As the other reviewer mentioned: the exceptionally long expressions should be broken up in some way for readability.

Out of the box I was unable to run the code as written for several reasons

1. example_run was renamed run_example
1. data_run was renamed data
1. The code is not python 3 friendly (print). I don't know what the policy is on compatibility with python 3 @oliviaguest any idea?

I made brief changes to allow the code to run and it did so. However plot_figures could not run as the naming convention in the data folder had changed. This needs to be addressed as I have not yet confirmed the plots match the article.

Lastly not at all important the comments for simpleModel have a variable number of tabs

# Article
As a comment the cross referencing seemed to have failed. I recall a similar issue and I put in explicit latex \ref and \label tags.

In general I felt the article was fairly sparse and could do with more text and commentary everywhere. Specifically in the introduction: a more general introduction was needed, what are the results, why are they important?

For the methods tables 1 and 2 should really be written into the text - they addresses several questions that arise while reading the text and they don't lend themselves to tabular form. Also the python versions are ?.?.? I don't know if that a cross-ref error or a place holder. Also the OS this was run on should be stated.

Lastly a sentence on the numerical methods is needed - nothing much just forward Euler and a statement (if true) that the results are unaffected by a smaller time step.

There are a few things in the results that need to be addressed. 
Firstly the summary of the main features should be presented in a clearer fashion... maybe this could be in a table?

Also in figure 2 (their figure 5) the frequency ranges are different. For example in the original 30Hz (controll) has a peak ~70 Hz but your implementation stops at 50Hz.

In the original paper power is in nA m^2 Hz^-1 can you scale your power measure to match? - you mention a scaling factor - have you tried to contact the original authors?

Also to match the original paper the raster plots should be flipped so that I cells are on the bottom.

All axis and tick labels need to be much bigger.

As for your comment about about being less pronounced, you proposed that it arises from the fact that there is a random process. What happens if you change the seed - does the effect size change?  In general this implementation does seem to replicate the original findings however, more work as to the nature of the discrepancy and its origin is required.

oliviaguest:
@rougier is there any policy regarding python versions?

rougier:
@oliviaguest @khinsen No official policy but a python 3 version would be better given that python 2 will be soon discontinued. Would a Python 3 version represent a lot of work in this case ?

aaronshifman:
@rougier I don't believe so, I think its just changing a few print statements to functions.

ChristophMetzner:
@aaronshifman @pietromarchesi  Thank you for your comments! I have already pushed an update which addresses some of your comments. In particular:
-	Added the additional argument to function rasterPlot in lines 55 and 56 of example_run.py
-	Removed obsolete analysis2.py script
-	Updated paths in run_simulations.py and plot_figures.py
-	Corrected indentations in simple_model_class.py
-	Changed print command to a Python 3 friendly print() (I haven’t worked with Python 3, so, is there anything else that needs to be changed to make the scripts Python 3 compatible?)
-	Corrected the version numbers in the script, i.e. replaced the placeholders with actual version numbers
-	Added a folder fort the results from averaging over trials
-	Cleaned up the scaling of the background noise rate
I am running some more simulations (in relation to some of your comments/questions) right now and will comment on those as soon as they are finished. I do still have an issue with the cross-referencing, I will make a comment as well, maybe one of you has encountered the same issue and might be able to help me.

ChristophMetzner:
@aaronshifman: 
Indeed the frequency ranges are different. This is simply a matter of displaying, which could easily be changed. However, I don’t think that it is of great importance here. The peak in the frequency spectrum in the original article you mention, is at 60Hz and is only a harmonic from the 30Hz peak introduced by the FFT. This is even more clearly visible in the plots for 20Hz drive, where, in the schizophrenic case, a 40Hz component is visible in the spectrum but the raster plot and the MEG signal do not show any 40Hz activity at all.

No I have not tried to scale the power in the frequency plots. In the original paper power for experimental spectra and spectra from the more biophysically detailed model are in nAm^2/Hz^-1, but not from the simple model I replicate here. The theta neuron model does not explicitly model potentials, currents etc. and thus there is no way to relate the resulting values to biophysical parameters. I mention a scaling factor simply because the raw MEG plots from the original article show peaks at around 0.15, whereas in my implementation peaks occur at around 5 or 6. I have not tried to contact the authprs regarding this scaling. This is because they did not reply to my initial email where I had asked if they were willing to share their code with me.

ChristophMetzner:
@pietromarchesi 
Indeed, in the absence of noise only a 20Hz component can be seen in response to 40Hz drive for the schizophrenia network (see below Figures, upper right panel). However, as can be seen in the lower left panels, the 40Hz component of the response to 20Hz drive in the control network completely vanishes. Furthermore, several small components are introduced in the response to 30Hz drive in the schizophrenia network (middle right panel).


replication-figure4-a-0-0
replication-figure5-a-0-0

I have run simulations with the noise amplitude scaled down to 75% and 50%, respectively. In both cases, the 20Hz component to 40Hz drive is increased, but the control network does not show a strong 40Hz component in response to 20Hz drive. Furthermore, at least for the 50% case, the 30Hz response does not match with the original article, similar as in the 0% noise case. Summarising, only a moderate reduction of noise (only applied to the schizophrenia network!) yields results closer to those reported in the original article, however, I don't think there is any evidence that noise is reduced in cortical circuits in patients suffering from schizophrenia (if anything it is the opposite).

(I can upload the Figures for the 75% and 50% cases as well, I just thought it might clutter up the page too much)

aaronshifman:
@ChristophMetzner I'm not sure I understand. Is there going to be another update, or is your update it for this round of revisions?

ChristophMetzner:
@aaronshifman I will push more updates in the following week addressing your other remarks/comments. Sorry for the confusion, I just thought I comment on some of the issues I had already fixed, instead of waiting for all of them to be addressed.

aaronshifman:
Not a problem: just wanted to clarify.

ChristophMetzner:
@aaronshifman @pietromarchesi I have just pushed changes that address the remainder of your comments/issues, i.e.
- I restructured the equations evolving the synaptig gating variables and thereby reduced the length and readibility
- I added more comments to the functions (e.g. getSingleSpikeTimes) and hopefully made it clearer what happens there.
- I added a sentence regarding numerical methods used. I also ran a set of simulations with a 4 times smaller dt, which gave the same results. This is also mentioned in the revised text.
- I increased axis and tick labels in all figures.
- I reversed I and E neurons in the raster plots.
- I summarised the main results in a table rather than a list.
- I incorporated the model description into the text and deleted the ugly Table 2. I kept the very brief model summary in Table 1, although it is now also described in the text, since I think Table 1 gives a nice, brief overview of the model.
- Regarding the choice of the startpoint for the calculation of the PSD: I used code I had written a couple of years ago and, honestly, can't remember what the purpose of choosing a startpoint other than the beginning of the array was. I therefore removed it without any observable differences in the results.

pietromarchesi:
@ChristophMetzner Many thanks for the new commits, I will review them in the next few days! Cheers

ChristophMetzner:
I had forgotten to update the notebooks, they are up-to-date now. Also fixed a minor bug in the restructured synaptic gating variable evolution.

pietromarchesi:
Here's a few more comments on the current state of the submission.
 
- With regards to Python 3 compatibility, note that Python 3 explicitly disallows mixing spaces and tabs when indenting, which is something that you do in your code. This is in general bad practice, as the indentation style should be consistent (without getting into the tabs vs spaces dispute), so I would endorse making the indentations uniform. 
- Indentations in the comments to the functions of `analysis.py` are also a bit off/random-looking, I would recommend fixing them (below the `def` line, the docstring should begin with only one indentation), and potentially conforming to the NumPy standard:
```
def example_function(param1, param2):
    """Example function with types documented in the docstring.

    Some more description if needed.

    Parameters
    --------------
    param1 : int
        A great parameter.
    param2 : str
        Another incredible parameter.

    Returns
    ---------
    int
        A magic number. 
    """
```

- Thanks for addressing my question on the effects of modulating the levels of noise, am I correct that the plot you showed is for the case of no noise at all? What parameter exactly where you modulating (so I can play around and tweak it a bit myself)?
- I still do not like (i.e. understand) the sentence stating that "the difference may only stem from a difference in noise". If I am not mistaken, the only stochastic effects are in the Poisson spike trains from the background noise, but didn't you just explain to me that changing that noise does not in fact bring you closer to the original result? Apologies if the question is trivial, but what do you mean exactly?
- I believe there is a mistake in the first bullet point when you describe the features of the schizophrenic network: the subharmonic component emerges at 20Hz, whereas the text says that it emerges at 40Hz.
- I find the labels in the current version of the .pdf to be still a bit small to be readable, definitely in figures from 4 onwards.

ChristophMetzner:
@pietromarchesi 
- I just noticed that I have not updated the figures in .pdf, I only updated the scripts to generate the figures. Will do that asap.
- I think the inconsistencies with the indentation come from the fact that I used different editors (on Windows and Linux). I will also fix that.
- Regarding the noise: Yes, the Figures I showed are for A=0, which means no noise at all. I agree that the statement 'the difference may only stem from a difference in noise', might be misleading or at best does not explain anything. I think it makes sense to include a paragraph on the effect the noise has. Regarding your point. that changing the noise does not bring the results closer to he original, well, this is true and not true. Changing the noise consistently (for the control and the schizophrenia network) does not bring the results closer, but, changing the noise only in the schizophrenia network does. I will include this discussion in the paper.

ChristophMetzner:
@pietromarchesi I have just pushed changes that address the figures in the article and the exploration of the effects of the background noise. The article now has a section where the exploration is discussed. Figure axis and tick labels are bigger now. Regarding the 'mistake' in the table, I think that it is not a mistake. The subharmonic component emerges at 40Hz drive and, therefore, is at 20Hz in the power spectrum. I will address the indentations within the next days.

ChristophMetzner:
Also, I could not resolve the cross-reference error when compiling the markdown file. Therefore, I removed the .md file, manually inserted \label and \ref commands in the .tex and compiled the pdf from the .tex file.

pietromarchesi:
@ChristophMetzner, I agree that "Regarding the 'mistake' in the table, I think that it is not a mistake. The subharmonic component emerges at 40Hz drive and, therefore, is at 20Hz in the power spectrum.", but it used to say that it was at 40Hz before your last few updates. It is all good now. 

Perhaps something will have to done about the formatting, there's many pages which now contain only one plot and are almost empty.

aaronshifman:
# Article

The article is much improved and cleaner compared to the previous version, however I still have a few issues, (in the order they appear)

- "are important for two reasons: First" - either colon and lowercase or two sentences
- "wide range of deficits in schizophrenic patients Second" - missing period
- "The neuron is described by a single variable \theta, which can be regarded as the membrane voltage..." - I think instead of treating \theta as membrane potential you could call it neuron state or oscillation phase.
- \frac{d\theta}{dt}
  - you index every other equation so \theta and S should have an index

- \alpha_j = \pm 1... \alpha_j isn't both it's either or, this should be clearer and represented in text or a piecewise function (but text would be clearer).
-you never define \eta or epsc (while it's a common acronym it should still be defined once).
- Should table captions be above or below the table, this seems like an editorial issue
- What do you mean by topology none (table 1), why is topology separate from connectivity.
-eta is missing from table 2
- you mention the scaling again, but you cannot verify the original code - have you tried to contact the original authors?
- crossref issue at fig:Vierling7
- I appreciate what you're trying to do with the exploration of discrepancies part - I feel like it needs a bit more explanation and clarification. Are you able to estimate how far off the reported parameters and the parameters you would "need" to match the presented results are?

# Code
The code as submitted does run with the exception of a typo (first comment)
In general the code is better commented and presents better, however there are a few more things

- Typo: line 23 of analysis.py there is no colon at the end of the function declaration
- The comment about drive current (line 141) should be in the text
- There is a large mix of tabs and spaces forming the indents, while Python 2 will still run this, it's not hard to change (a single auto-format in pycharm will correct it), and I believe after that it becomes Python 3 compatible
- Variables are duplicated it appears (for example n_ex is defined as 20 in example_run as well as the default in simple_model_class.py). If it's constant leaving it as a default is fine, but remove the redundant declaration.
- You have some unused variables ex. line 354 of simple_model_class.py
- Many lines are still far too long

In general while observing pep8 to the nth degree may not be necessary certain things like max line lengths are really important for readability.

# Figures
- There is a lot of duplicated code in both plotting scripts, and this might be my personal preference bleeding in but I would prefer to have common plotting code as functions with parameters this (a) makes it cleaner and (b) makes it so that changes are far easier because I do have some requests for those
- Figure 1 looks really good the only issue I have is that the matrix labels overshadow the axis labels by way too much and the axis labels are still too small. What I think would work is to have the two sets of labels the same size and leave the matrix labels bold.
- the same thing applies for the tick labels, you could make fewer and make them a bit larger
- I have the same comment for the rest of the figures, shrink the matrix labels and make the axis labels larger.
- For figures 3-6 if you cut them by a bit (maybe 25%) you might be able to fit more than 1 per page and the paper would look far less sparse.

aaronshifman:
Thank you @ChristophMetzner, I'll take a look this weekend.

ChristophMetzner:
@aaronshifman I have not yet addressed the Figures, still working on that.

ChristophMetzner:
@aaronshifman I have just updated the plotting scripts and notebooks.

ChristophMetzner:
@aaronshifman Regarding your other comments:
- "What do you mean by topology none (table 1), why is topology separate from connectivity."
By topology I simply mean the arrangement of cells, i.e. are they arranged in a 2D or 3D space having coordinates. While this would be better called geometry, many articles use the word topology (also reflected in the names some simulators use, e.g. NEST has a topology module which arranges cells in 2D or 3D space). In this network there is no need for cells being arranged in space, so there is no 'topology'. I guess the difference to connectivity is also apparent now. 
- "you mention the scaling again, but you cannot verify the original code - have you tried to contact the original authors?"
I had contacted the authors in the beginning, about sharing their code in general, without getting aresponse. Therefore, I did not contact them again asking about the scaling.
- " I appreciate what you're trying to do with the exploration of discrepancies part - I feel like it needs a bit more explanation and clarification. Are you able to estimate how far off the reported parameters and the parameters you would "need" to match the presented results are?"
Unfortunately, I don't see a way to estimate how far off I am, given that we have this highly nonlinear system. I also see various other ways by which the discrepancies could be explained. For example, a 'small' change in the weight from I to E cells (an increase) would probably lead to a more pronounced 20Hz component for 40Hz drive in the schizophrenia network (since the total inhibition would be increased, and therefore, more E cells would be suppressed eevry other cycle). However, I think that this would again reduce the 40Hz component in response to 20Hz drive in the control network, similar to the noise exploration. I can explore this further, however, I am not sure what we will gain from that, since there are more possible ways than the one described above. A complete and thorough search of the parameter space, I feel, is a little bit beyond the scope of this replication, since the main features are being reproduced and only the effect being slightly reduced. Let me know what you think. @pietromarchesi What do you think?

ChristophMetzner:
@aaronshifman @pietromarchesi  Now there are most of the time two figures per page (except for Figure 1 and Figures 7 and 8). Should I make 7 and 8 smaller so that they fit on one page?

pietromarchesi:
@ChristophMetzner Thank you for your last commits and for your comments. Regarding the last point you raise in your comments, on exploring the discrepancies, I agree with you, and, although interesting and welcome, I do not find it strictly necessary to investigate further possible reasons for discrepancy, given that the main results have been replicated. 

I think two figures per page is nicer, if they could fit nicely and be readable. I still find it suboptimal to have several pages of figures that completely break off the text, but I don't have an immediate solution for that. 

I have one more tiny aesthetic comment: in `run_main_simulations.py` (and perhaps elsewhere), when lines are broken up then the second part of the line that falls below should be indented. 

I will be on holiday and with limited access to internet from next week. My overall recommendation at this point in time is that the replication (with a few smaller edits) should be accepted.

aaronshifman:
@ChristophMetzner Thanks for the commits. I hope to get back to you within the coming week.

aaronshifman:
Hello @ChristophMetzner, I apologize for the delay in the review. I appreciate your addressing my comments and I have no further scientific issues. 

My only issues are a few typos / omissions

- page 1: schizophrenic **patientas**
- page 1: **loosing** much of its simplicity
- page 11: features **a re a** little bit
- page 2, second equation: `S_k = \Sigma_j^n ...` : n isn't defined in the text (I assume its over all neurons, just a sentence in the text to define it would help)

I agree with @pietromarchesi that the disjointed figures aren't the best option however I don't have a better option and it seems to be that the style that the figure layout isn't nearly as important as the replication itself.

Given this, I move to accept this paper (conditional on these small editorial changes).

ChristophMetzner:
@aaronshifman @pietromarchesi  I have fixed the typos/omissions and have rearranged the figures and the text and I think that the layout of the article is ok now.

rougier:
@oliviaguest Anything more needed ?

oliviaguest:
I do not think so. Will move to publish it now.

oliviaguest:
Done! :smile:

[DOI](https://doi.org/10.5281/zenodo.854617)

PR #33, opened 2017-06-20
Review Request A. K. Diem

alexdiem:
**DIEM**

Dear @ReScience/editors,

### Original article

**Title:** A bidirectional model for communication in the neurovascular unit  
**Author(s):** Witthoft A, Karniadakis GE
**Journal (or Conference):** Journal of Theoretical Biology
**Year:** 2012
**DOI:** [https://doi.org/10.1016/j.jtbi.2012.07.014](https://doi.org/10.1016/j.jtbi.2012.07.014)
**PDF:** [http://www.sciencedirect.com/science/article/pii/S0022519312003487](http://www.sciencedirect.com/science/article/pii/S0022519312003487)

### Replication

**Author(s)**: Diem AK
**Repository**: [https://github.com/akdiem/ReScience-submission/tree/Diem-2017](https://github.com/akdiem/ReScience-submission/tree/Diem-2017)
**PDF**: [https://github.com/akdiem/ReScience-submission/blob/Diem-2017/article/Diem-2017.pdf](https://github.com/akdiem/ReScience-submission/blob/Diem-2017/article/Diem-2017.pdf)
**Keywords**: neurovascular unit, neurovascular signalling, neurovascular coupling, astrocyte signalling
**Language**: Python 3.5.3
**Domain**: Neurovascular coupling

### Results

I believe the original results have been almost fully reproduced as explained in the accompanying article.

---

**EDITOR**

* [x] Editor acknowledgment `June 28, 2017`
* [x] Reviewer 1 started (@ChristophMetzner) `July 4, 2017`
* [x] Reviewer 2 (@TiinaManninen `July 7, 2017`
* [x] Review 1 decision [accept] `September 6, 2017`
* [x] Review 2 decision [accept] `August 31, 2017`
* [x] Editor decision [accept] `September 6, 2017`


rougier:
@akdiem Thanks for your submission, we'll assign an editor soon. Do you have a link to an open-access version of the original article ?

rougier:
Dear @ReScience/associate-editors 

Can someone edit this submission (domain is Biology / Neurovascular coupling)?

alexdiem:
@rougier The original authors have a pdf on their website
 [https://www.brown.edu/research/projects/crunch/sites/brown.edu.research.projects.crunch/files/uploads/Alix_bidirectional.pdf](https://www.brown.edu/research/projects/crunch/sites/brown.edu.research.projects.crunch/files/uploads/Alix_bidirectional.pdf) I didn't upload it into my repository because I did not want to infringe on the journal copyright.

rougier:
Thanks, a link is perfect !

alexdiem:
Is it acceptable to make small changes to the manuscript at this point? I have just realised that I forgot to add a detail for implementing one of the equations, which I only got to know about by contacting the original authors.

rougier:
No problem, you can

rougier:
@akdiem Just to let you know I did not forget you, an editor will be assigned soon hopefully.

rougier:
@benoit-girard Can you edit this submission ?

benoit-girard:
Yes

benoit-girard:
@ChristophMetzner Do you think you can review this submission?

ChristophMetzner:
@benoit-girard At first glance it seems a little bit out of my comfort zone. I will have a closer look this evening and reply tomorrow the latest.

ChristophMetzner:
@benoit-girard Had a closer look and, although I am not an expert on the topic, I feel confident to review.

benoit-girard:
@ChristophMetzner OK, great, thank you. 
It is difficult to find someone fully comfortable with the topic, given our current set of reviewers. I have been contacting potential reviewers by mail, without much success up to now. I am waiting for answers, that's why we will have to wait a bit before we have a second reviewer.

benoit-girard:
A second reviewer is expected to join soon. Thanks for your patience.

alexdiem:
@benoit-girard Ok, thanks for the update!

benoit-girard:
@TiinaManninen will be the second reviewer (Tiina, can you confirm here that you accepted the duty?).

rougier:
@benoit-girard Don't forget to update relevant dates at the top, you'll need these information later.

TiinaManninen:
I will accept the duty.

benoit-girard:
@ChristophMetzner @TiinaManninen A simple reminder: here are reviewer guidelines: https://github.com/ReScience/ReScience/blob/master/reviewer-guidelines.md

TiinaManninen:
Dear @akdiem and @benoit-girard,

I have now reviewed the present versions of the manuscript and the model code. In the review, the following notation was used:

Witthoft and Karniadakis (2012) model = 2012 model 
Witthoft et al. (2013) model = 2013 model

Please, find below the suggestions separately for the manuscript and model/simulations.

In the manuscript:

a) Introduction of the manuscript states that there is bidirectional signaling between neuronal and vascular activity in the 2012 model. However, this is not true. The bidirectional signaling happens in this model only between astrocytic and vascular activity. See Figures 1 and 2 in the original article for more details. The 2012 model does not include modeling of actual neurons but instead includes glutamate in the synaptic space as stimulus (glutamate in the synaptic space is a smooth pulse approximation thought to be released from active neurons) and K+ in the synaptic space as variable (K+ in the synaptic space depends on the other stimulus which is the smooth pulse approximation of K+ that is thought to be released from active neurons). Synaptic space means the extracellular space in the vicinity of the pre- and postsynaptic neurons, not the intracellular space of the neurons. Note that K+ and glutamate in synaptic space do not depend on the other variables of the model, thus there is no feedback from the other cells of the model to K+ and glutamate in the synaptic space.

b) In the manuscript, please give Equation 31 without "_" and "$" thus not showing it as a Latex code in the pdf.

c) How do you know that the original authors used [Ca2+]_ER=400 uM in the 2012 model because the value was not given in the original article?

d) Please, modify "an equilibration phase (t = -20 - 0s) without stimulus presentation had to be implemented." to "an equilibration phase had to be implemented for a length of 20s before the actual simulation." or to something similar.

e) Please, explain in the manuscript why rest of the original figures were not reproduced. It is possible that when trying to reproduce the other figures will actually help in reproducing better the results of Figure 5 of the original publication.

f) Please, delete extra ")" in "(K+))" in Result section and in figure text of Figure 1.

g) Please, correct "publcation" to "publication".

h) Should "or indeed 1" be "or indeed in Figure 1"?

i) Please, explain in the manuscript why is it impossible to use the code of the 2013 model to fix the new implementation of the 2012 model.

In the model/simulations:

j) The code is understandable (but not commented that much). I was able to simulate the model and see the plotted results with jupyter notebook. However, I got the following warning: ../anaconda/lib/python3.5/site-packages/ipykernel/__main__.py:251: RuntimeWarning: overflow encountered in exp.

k) Please, include the code and data file of V_K to reproduce Figure 2 of the manuscript.

l) The 2012 model does not give details how I_Ca current in Equation 36 was changed to J_Ca flux in Equation 23 but the model has 1/(2*alpha)  I_Ca as flux in Equation A.1 that you modified to alpha  I_Ca. Do you think that the latter one should be used as J_Ca in Equation 23, thus J_Ca = - alpha  I_Ca and not J_Ca = I_Ca/(C_smc  gamma)? When you test this, then please make sure the signs in Equation 23 and A.1 are correct. 

m) Please, explain what do you mean by "sinf is inconsistent between code and equations, gammacai is off by order of magnitude" and by "why *cm here? Is that the artery length?".

n) Please, explain in the manuscript how you obtained the values for Max_neural_Kplus and Max_neural_glut because these values were not given in the 2012 model. In the 2013 model, they were 13mM and 0.75, respectively.

o) Please, change your stimulus as in the 2012 model: the stimulus should be between 5s and 25s as in Figure 5 of the original publication, not between 2.5s and 18.75s. After this change it is much easier to see how well the new results match the original results. At the moment, the results do not match because the curves drop down too early because the stimulus ends too early. Please, give the details of the stimulus in the manuscript.

p) Is Equation 40 used instead of Equation 27 in Figure 5 of the original article? Please, test.

q) Should rho multiply only I_Ca and not [Ca2+]_SMC in Equation A.1? Please, test.

r) The equations in the 2012 model and the 2013 model do not seem to be that different. I would suggest to look in more detail the 2013 model, the supplementary of it, and the code they provided. With this knowledge, it might be possible to fix the new implementation to reproduce results that are closer to the original results of the 2012 model. Some ideas to test below.

s) The last bracket in Equation 5 in the 2012 model is after J_TRPV but in 2013 model it is after J_leak. Please, test how this affects the results.

t) J_TRPV = -1/2  I_TRPV/(Cast  gamma) in the 2013 model compared to J_TRPV = -I_TRPV/(Cast * gamma) in the 2012 model. Please, test how this affects the results.

u) Unit for k_Ca is different in the 2013 model. Please, change the value to 135.68 1/s so the units match in the corresponding equation of the model 2012 and give the details in the manuscript.

v) The supplementary of the 2013 model shows the astrocytic currents (Figure S.2). Please, plot your currents (give that version of the model code for the reviewers) and compare your currents to the currents of 2013 model. Hopefully this will help in seeing if it is possible to obtain astrocytic membrane voltage that is closer to Figure 5 of the original article.

w) Please, make sure you give all the model details in your manuscript that were not given or were given wrong in the original article.

alexdiem:
Dear @TiinaManninen,
thank you for the thorough and detailed review. As I am implementing the changes into the manuscript I have a question regarding the citation of correspondence with the authors of the orignal paper. Perhaps this requires and answer from @rougier or @benoit-girard. What is the best practice when writing about personal correspondence in the article? I have a long email trail with the original authors, some of which was vital to spot errors in the manuscript but another is crucial to the reviewers request (i): I attempted to revert the code I obtained from Witthoft to the 2012 version, but could not obtain the same results, so I emailed Witthoft about it who stated that it is not surprising to her that I cannot revert the code back to its previous version because of the sensitivity of the parameters, of which a substantial number would need to change. That process would be hugely error prone. How should I best handle this?

benoit-girard:
@rougier: what do you think about @akdiem question on correspondence with original authors?

My point of view is that, as ReScience explicitely encourages contacting the original authors when having difficulties to reproduce the work, you should explicitely mention that you contacted, and summarize the content of this interaction, more or less as you did in the above comment.

benoit-girard:
@ChristophMetzner do you think you could finish the first round of review before the end of this week? If not, when do you thing it is feasible?

ChristophMetzner:
@benoit-girard I am at a conference until the end of the week, so I think it  is unlikely that I will finish the review this week. I will try to finish it as soon as possible next week.

benoit-girard:
@ChristophMetzner OK, that would be the 28th, then.

rougier:
Correspondence with original authors is actually encouraged if there's a problem in the replication. For example, one (of the many) goal for a replication is also to collect missing parameters and to display them in the new article.

alexdiem:
Dear @TiinaManninen @benoit-girard 
Please see below for my responses to the review by @TiinaManninen:

a) This is correct and the introduction has been modified accordingly.

b) The closing "$" was missing, hence the equation was not properly displayed in the pdf. It has been added.

c) The original publication states that the equations containing [Ca2+]_ER were obtained from Bennett2008, which included a value of 400 uM. A statement to reflect this has been added to the list of modifications in the paper.

d) The phrase has been modified as per the reviewer's suggestion.a) introduction modified

e) I will attempt to reproduce Figures 3 and 4 in Witthoft2012, this will require some additional time.

f-h) The typos have been corrected.

i) An explanation has been added to the Results section.

j) The overflow warning happens at beginning of the equilibration phase and it is the reason for requiring the equilibration of the model. Thus, to avoid confusion, I have suppressed the warning in the code.

k) The data for V_k in Figure 2 was already included in the data directory. I have added the code for the simulation to the repository.

l) Indeed, details of how J_Ca was defined were not given. Correspondence with original authors suggested a definition analogous to J_TRPV, but using C_smc instead of C_astr. I tested the reviewer's suggestion (and also Jca=alpha*Ica), but neither change makes a difference to the results. The problem with a highly nonlinear system like this one is that it is possible, that a small change on its own might not make a significant difference. Instead, the correct combination of changes has to be found, which, given the number of possible combinations, can be impossible.

m) These comments were questions I had while reviewing the code for the 2013 model. Because they are confusing I have removed them.

n) A Figure showing the input together with an explanation of how the values were chosen was added to the manuscript.

o) This has been changed in the code and the figures have been updated.

p) Using eq. 40 instead of 27. leads to fast oscillations in SMC Ca2+ and arterial radius, similar to the inset window in Figure 4 in the original publication. Thus it is more likely that equation 27 was used.

q) This change leads to a linear increase in SMC Ca2+ for the entire simulation time and the results for the radius become nonsensical.

r) This is what I have attempted to do over the past two months in continuous correspondence with the original authors. Nominally the equations do not appear that different, but because the system is highly nonlinear small changes in the equations can lead to large changes in the results, making it difficult to backtrack. Because this has only been successful to the point that the results match apart from Vk has lead me to decide to submit at this stage and obtain the reviewers' input.

s) I have tested this previously. This change makes some difference to the astrocytic Ca2+ and EET such that these values do not reduce to zero at 30s, but only at 50s. Between 30-50s the values are approximately a quarter of the value of the flat part between 15-30s.

t) This does not change the results noticeably. See also my additional comment to (l).

u) This change is already implemented in the code, but the article incorrectly still listed the units for k_Ca as nM/C even though it should be 1/s.

v) Unfortunately, the values for the astrocyte currents are not very comparable between the 2012 and 2013 versions of the model. However, I have added the corresponding plot and and explanation to the Results section of the manuscript.

w) I believe all changes are now reflected in the manuscript.

Please let me know whether these responses and modifications to the code/manuscript adequately answer your requests. Thank you again for the detailed review, it is very valuable and useful.

alexdiem:
I have done some more testing with attempting to reproduce Fig. 4 in the original manuscript. This revealed two more changes to equation (23): It should be defined analogous to (22), currents have positive signs and are divided by the respective volume ratios. I have added three more plots to support this. The first shows (23) with positive signs but without dividing by volume ratios:
Pinacidil without dividing currents by volume ratios 
Radial strain looks similar, but the curve's increase is late compared to the original figure. Ca2+ does not match. Dividing currents by the volume ratios like in equation 22 gives:
Pinacidil with dividing currents by volume ratios 
The shape of the curve matches better, but the values for Ca2+ are off by quite a bit. The 2013 model uses  Ca_p0 = 5 uM instead of 2000 uM, so I tried to use that:
Pinacidil with Ca_p0 = 5 uM 
This changes the shape of the curve, so it seems that Ca_p0 = 2000 uM would have been the value used in the 2012 paper. The fact that the figures still don't match, however, means that there must be further changes to be made.

ChristophMetzner:
@akdiem Thank you for the effort you have so far put in to addressing all the useful comments from @TiinaManninen. I have finally had a chance to look at the submission. Here are some comments on the manuscript from my side. I am currently looking at and running the code, I will comment on the code within the next couple of days. 

### Manuscript 

a) Mentioning simulation tolerance values in the Figures is enough, don’t need them in the text (or if at all, in brackets). They disturb the reading flow a little.

b) Please explicitly state the input function in the article.

c) Please write parameter K_Ca as 135.68 instead of 1.3568e2, since that is easier to read.

d) Page 2, last paragraph: ´…and the solution values at the end oft he simulation *are* 
     (instead of *is*) used…´.

e) Typo in the figure legend of Figure 1, last sentence: should read *original* (instead of *origial*).

f) Please describe more detailed what you mean by ´the shape of V_k, which in the first half looks very different…´. What exactly are the differences? Can this tell us something about where the error might lie?

g) At the end of the results section you look at I_sigmaK and I_TRPV and match it to the newer implementation. Maybe I didn't completely understand, but what happens if you scale I_sigmaK and I_TRPV so that they match? How does this affect the results? 

h) I think that it would be very useful to include a box or table, where you state all the equations that are used in the implementation of the model. I know that they are already in the original, but you have found many errors or incomplete definitions/statements, that it would be good to have all *correct* equations in one place.

i) I think you have done a great job in hunting down all errors, missing definitions and parameters, etc. The only thing that is missing, in my opinion, is a discussion of the plausibility of the results produced with your code. Do the results look as one would expect? As you have already pointed out, tracking down all differences between the original model and your implementation has become a very hard task. Now we have to judge whether it is a necessary one. If your model implementation gives results that are more plausible, I think we do not need to do so anymore.

TiinaManninen:
Thank you @akdiem for addressing the issues I named earlier. I also agree to all the issues provided by @ChristophMetzner, and especially since more errors have been found, you should give in tables the original and corrected equations side by side.

I read the new version of the manuscript and here are few points that need to be corrected.

a) Please, remember to run bibtex too because it seems that the references are not in the order of appearance in the reference list.

b) When citing multiple references, please modify [6],[4] to [4,6] (note the increasing order).

c) Please, give the values without "e" in the manuscript.

d) You give several points in section Methods. Please end every point with a period.

e) Code has J_Ca = I_Ca/ (C_smc * gamma) but manuscript has J_Ca = - I_Ca/ (C_smc * gamma). Please, make sure that all fluxes have correct signs.

f) J_TRPV cannot have positive sign both in original Equation 5 and in your version of Equation 23. It should have negative sign in Equation 23. Please, make sure that all your fluxes have correct signs.

g) Please, modify "Table B2" to "Table B2 in [1]", so always give the reference too.

h) What do you mean by k_Ca incorrectly cited from [6]? Please, specify in the manuscript.

i) Please, modify "IP3" to "$IP_3$" in the manuscript.

j) In the figure text and in Figure 1, please change the unit from uM to uM/s.

k) Sometimes the citations in the text are with black color and sometimes with blue color.

l) The original concentration for extracellular K+ in the perivascular space has max value of 20 mM (Original Figure 5), but the reproduction has max 15mM (Figure 2). Please, try to explain the difference in the manuscript.

m) Please, modify "defintions" to "definitions".

n) In your Figure 5, the maximum voltage is about -45mv, not -25mV as you give in the text. The original minimum value is about -75mV and not -80mV as you state in the manuscript.

o) Please, run also the simulation without TRPV4 channels as in original Figure 5 and plot it also to your Figure 2.

p) Please, give details in your manuscript which code files are used for which figure.

q) Please, give your version of the original Figure 4 in your manuscript.

I will look at the code later, after @ChristophMetzner have given the comments and possible corrections are done.

ChristophMetzner:
@akdiem I had a look at your code. I could run all four different 'main' scripts, and, except for 'main_trpv.py', the plots were produced correctly. In 'main_trpv.py', the _plt.show()_ command was commented out. After uncommenting this, the plots here were also produced correctly.

The main issue I have with the code is that it is very sparsely commented and, although the code is understandable, I think more comments would greatly improve the (re-)usability. Therefore, I suggest the following:

- a) Include a header in every file which states the author, the Rescience publication and a short description of the content of the file.

- b) Include docstrings for methods (e.g. according to the numpy standard below):
```def example_function(param1, param2):
    """Example function with types documented in the docstring.

    Some more description if needed.

    Parameters
    --------------
    param1 : int
        A great parameter.
    param2 : str
        Another incredible parameter.

    Returns
    ---------
    int
        A magic number. 
    """
```
- c) Add more comments within each method. Briefly describe what each line (or each couple of lines, if they are linked) does. Try to link to the equations from the article.

- d) Since there are now 4 main scripts, please update the _Usage_ section. Briefly explain what the 4 scripts do and how they are used.

alexdiem:
Sorry for the delay in getting back, but I had some other papers to finish to deadlines. This is my response to the comments by @ChristophMetzner to the manuscript:

(a) I've removed the tolerance values from the text

(b) I don't think that value is added by explicitly stating the input function in the manuscript. It is a simple step function with fixed values, which I believe is much easier to read from the code. However, I have added the maximum values for the input function to the manuscript.

(c-e) I modified the manuscript accordingly

(f) I added a description of the differences in V_k. The problem with the differences is though that the equations are highly nonlinear, so any changes made in any of the equations can have an impact anywhere in the model. This is why it is so difficult to figure out exactly what might still be missing from the descriptions in the original publication. 

g) The problem again lies with the high nonlinearity of the model. Rescaling the two variables by 1/100 essentially removes the vasodilation response of the vessel Ca2+ channels and vessel radius.

h) I have an agreement with the original authors that we shall write a joint Letter to the Editor with the full corrected model once all the corrections have been identified.  

i) Thank you. I find it difficult to judge whether my result for Vk is more plausible or not than the one of the original publication, but I am looking for experimental publications to see what it should look like. So far that has been unsuccessful, but I will try again.

alexdiem:
My response to the new requests for the manuscript from @TiinaManninen:

a) Bibtex is run by the tex template from the journal, I have no influence on the order of the citaitons

b) Again, this is from the journal tex manuscript and I have no influence in the appearance of the citations

c) I believe k_Ca was the only one and I have modified it

d) I have corrected the bullet point list

e) I have corrected the manuscript. I've also tested both signs in the code again, and interestingly, it does not make any difference whether J_Ca has a positive or negative sign.

f) It is correct that J_TRPV and J_Ca cannot have positive signs in equation 23. This has been corrected in both the code and manuscript.

g) I changed the manuscript accordingly

h) It means that the original authors cite a source for their value, but the value they state in the table does not match the value in the source publication they cite. What would be a better way of phrasing this?

i) I changed the manuscript accordingly

j) I corrected the figure axis label.

k) I couldn't find an example for it. Do you mean equation labels? That is because they don't refer to an equation in my manuscript (so it can't have a hyperlink). Hyperlinks (blue) are provided wherever I can link them to a part of my manuscript.

l) I cannot explain where exactly the difference comes from (again, this is an issue with the nonlinearity of the model and the reason why it is such a struggle to fully reproduce it), but I have added it to the manuscript.

m) I have corrected the typo.

n) I have corrected the range of values in my manuscript. It should have read -60 to 50

o) The original figure 5 contains the TRPV channels, so I do not know what to compare the model without TRPV to. However, I have run the model without +J_trpv in equation 5, -J_trpv in equation 20 and J_trpv in equation 23. This does not appear to make a difference in the results:
NVU model without TRPV channel 

p) I have added these to the manuscript

q) I have added the figure to the manuscript.

alexdiem:
I have added the docstrings and headers as requested by @ChristophMetzner. Note however, that I have explicitly not added documentation for every single parameter in nvu.py as I believe this would make the code a lot less readable as much more scrolling would be required to find a certain section. Instead, I added a comment stating which equation is being implemented for each line. This (plus having named all parameters in accordance with the parameters in the original publication) should make it obvious enough which parameter in the code refers to which parameter in the original publication.

benoit-girard:
@ChristophMetzner @TiinaManninen Please let us know, as soon as you can, whether you are satisfied with the enhancements provided by @akdiem or have additional requests.

TiinaManninen:
Thanks for addressing the issues. The original Figure 5 has simulation results from both with the TRPV channels (black color) and without them (gray color). I will check the manuscript and the codes this week.

alexdiem:
Ah I did not see the grey dashed lines, they didn't come out properly on my print out! I've also just resolved the difference between my and the original astrocyte Ca2+ and EET plots - if the input function increase is a tanh function I can reproduce those. Still get the same overshoot on Vk, but I'm going to work this into the manuscript.

TiinaManninen:
I checked the manuscript and the code earlier today, but now see that @akdiem have modified both. Please see below my comments, and correct them if not yet corrected. Furthermore, let us know when you have finalized your manuscript and code, so we can check them once more.

Comments to the manuscript:

a) Introduction, 3rd paragraph: "The paper contains" -> "The paper [1] contains".

b) I think there has to be something to get the references numbered by the order of appearance, because in your first version of the manuscript they were like that. 

c) It would be nicer to have, e.g., "[7][5]" -> "[5][7]".

d) Please, do not use, e.g., "1e-4" but use "$10^{-4}$" in the manuscript (atol, rtol).

e) Please, explain the abbreviations the first time they appear in the text, e.g., Glu, K+, SMC, EET. 

f) Please modify the unit for glutamate in the figure text of Figure 1 from "uM" to "uM/s".

g) I would not say that the original and new K_psv results match, so please modify the sentences "Potassium (K + ) in both the synaptic and perivascular space match their corresponding plots in the original publication, as do calcium (Ca 2+ ) in the smooth muscle cell and the arterial radius" and the figure texts. For example, you could say for K_psv that the shape of the curves matches but the actual values not.

h) In this sentence, the citation is with black color, and should it be [2] instead of [6]: "It is unclear, where this difference comes from as [6] does not indicate that any of the equations affecting Ca 2+ or EET have been altered."

i) In "d V_k / d t", V_k is in italic, but otherwise not.

j) Please write as "(Equation 2 in [citation])" in "sigK is mostly defined by the values for potassium in the synaptic space (equation 2)", and similarly for "I_BK (equation 15)", "equation 2", and ("V_m, equation 30)".

k) Please explain more carefully, what potassium values you mean: "Because there is a discrepancy of the potassium values by three
orders of magnitude between the two publications by Witthoft et al.,"

l) You could modify "paper" and "manuscript" always to "article" or "publication".

m) Please, give k in subscript in "They obtained data of the astrocyte membrane potential (Vk)". 

n) In the figure text of Figure 5, please explain which curve is which.

o) Please, modify "-72 mV to -40 mV" and "-80 mV to -40 mV" to "-75 mV to -40 mV".

p) I don't see these values in Figure 5 "while it is approximately -60 mV to 50 mV here", so please modify the text.

q) Please add "Figure" to "The results of this modification (main_k.py) are shown in 6.".

r) Please, explain in the text why Figures 6-9 of the original article were not reproduced.

Comments to the code:

a) Your manuscript and code have different values: the code has Ax = 50.265 * um**2.

b) In nvu.py, calcium_p_dt = - Jca/VRps - Cadecay * (calcium_p - calcium_p_min). Why Jtrpv is missing?

alexdiem:
It appears that J_trpv also has the wrong sign in the original manuscript. Plotting both with and without TRPV channels I get the following results with J_trpv having a negative sign (original manuscript):

J_trpv with negative sign

with a positive sign I get

J_trpv with positive sign

which seems a better match

alexdiem:
@TiinaManninen is this now your final set of comments? It is of course ok to check again with comments you have previously made where you believe my answer wasn't sufficient, but every submission also contains entirely new requests, which is just a little bit frustrating.

alexdiem:
@TiinaManninen Please see below for my responses. I trust that this is the final set of comments. I apologise that in the process of responding to the reviews of this manuscript I have found more modifications to the model that get my solutions closer to the original solutions

a) modified

b-c) Orignally I hard coded the references because I didn't know how the references worked. So no, it's out of my hands.

d) Because this is just specific to the code I think it is better to write it exactly as in the code and will leave it as it is.

e) Modified

f) Modified

g) K+ in the synapse matches in my eyes. The figure in the original publication looks just a bit more stretched but looking at the maximum value and the time points, over which increase and decrease happen, they match. The maximum value of K+ in the pvs is slightly lower. So I maintain what I have written in the manuscript.

h) Modified. Citation was overseeen when changing from hard coded to bib file usage.

i) Apologies, but I couldn't find the location of the issue

j) I think this should be clear from the context of the manuscript and just adds unnecessary text. 

k) I have added "synaptic" to the manuscript.

l) I could, but I don't see an added value to going back and modifying them. 

m) I think this is better written the same way as in the code and plots, hence it is Vk.

n) Modified.

o-p) Apologies, I was looking at the original fig. 5 and my fig. 2 by accident when comparing values.

q) Modified

r) I believe all relevant plots for showing to what extent the model can be reproduced have been included.

ChristophMetzner:
@akdiem Sorry for taking so long, I will give a (hopefully) final reply tomorrow.

ChristophMetzner:
### Code
a)	In nvu.py synapse() input is described as a step function, however, in the newest article version, input rises with tanh?
b)	I appreciate the reference to the equations from the original article and I agree that documenting all the parameters/variables might not improve readibility/understandability. The documentation of the key parameters together with the reference to the equations is probably the best option.
c)	However, one minor detail to the referencing: I think simply writing eq 1, for example, does not stand out enough from the code and makes it thus hard to find the implementation of specific equations. Therefore, I would change it to Equation 1 for example.
d)	The original article states J_BK=I_BK/(Castr*gamma), but shouldn’t this also be negative since it is a flux of positive ions? Same for J_KIR.
e)	I tried to play around with the code again, and it didn’t work. The loading of nvu did not work. After adding an empty _ _ _init_ _ _.py in the nvu folder, the import worked. Please fix that.
f)	Running any of the main scripts now gives a syntax error in nvu.py in line 361. Please fix that, I would like to be able to play around with the code.

### Article
a)	Despite the agreement to write a letter to the editor, I still think that this article should contain a comprehensive list of all equations that are implemented in the code! I think that the number of errors already discovered in the original manuscript, emphasize the need for such an overview. Furthermore, the code should then refer to the equations in this list. It is one of the main purposes of the ReScience initiative to provide reimplementations that are not only free and useful, but also that allow to understand the implementation. A list of all equations is absolutely necessary for this.
b)	Minor detail: When you list all changes to the equations, you write that the value of g_trpv was incorrectly cited from the source and that you found the correct value. Please explicitly specify the correct value.
c)	Figure 2 does not explain what the dashed black line is.
d)	Figures 5 and 6 do not reference to the Figure in the original publication (Figures 3 and 4 there), please add.
e)	In Figure 6, although it is very hard to see, it seems that radial strain starts at 0 in the original figure, but at 0.025 in your figure. Anyway, the general shape matches, however, not only the onset of the rise comes later in your plot, the amplitude of the fluctuation seems to be bigger too. PLease also mention this.
f)	Could you also provide a zoomed in view, like in the original article, to check that the overall behaviour ist he same.

alexdiem:
@rougier Your request (a) will amount to an almost complete copy of the original paper and it worries me that infringement of a journal's copyright should be suggested and explicitly requested here. @benoit-girard

ChristophMetzner:
@akdiem I disagree, I don't think using equations from an article (correctly cited of course) is copyright infringement . If that was the case, one could not use any equations ever published. But let's wait what @rougier and @benoit-girard think.

TiinaManninen:
@akdiem I fast read your manuscript and I am kindly asking you to think few minor details in your writing:

a) It seems that some abbreviations are given in the text but then the abbreviations are not always used later in the text but the actual words are. On the other hand, word "calcium" is used first time in introduction, line 15, but the abbreviation is given in Results section. Abbreviation EET is explained in one of the figure texts, please explain it in the text in Results section. So, please make sure all abbreviations are explained in the correct place.

b) Sometimes you give units in italics and sometimes not. I would recommend not to give them in italics.

c) my old i) where I meant the partial derivative on page 5, line 7:  In "d V_k / d t", V_k is in italic, but otherwise not.

d) Since Ca_ast and EET in Figure 2 now match much better the original results (there is no more the large overshoot in the first peak), the result sentences should be modified in the text and in the figure text. If there still is difference in the first peak compared to the original result, then give how big the difference is.

e) I agree that the manuscipt would be much clearer if all equations are given. If possible the original and new equations could be given side by side.

Let us know when you have fixed the manuscript and the code, so we can make sure the code produces the figures in your manuscript.

rougier:
I doubt that reporting equations in the new manuscript would be a case of copyright infringement. I think it may fall within the scope of fair use. One potential solution would be to reference a preprint (CC-BY) and we would be on the safe side. But I'll try to look for information (or @khinsen, do you know the answer?)

alexdiem:
If I copy all equations into my manuscript then it will also require the explanations of the equations and that would quite frankly be a complete copy, a paraphrased one, but paraphrasing does not protect from copyright infringement. This also has to be considered more carefully than "I think". 

In terms of what I could read as the scope of the journal I have adequately described my code (par the last few corrections I have to look at), including exactly where it differs from the original paper and the code is openly accessible. Also, from your instructions to authors: 

> If both reviewers managed to run the code and obtain the same results as the ones advertised in the accompanying material, the submission is accepted.

and also from the reviewer guidelines: 

>The role of the reviewer is thus to verify it the proposed submission is actually replicable. This means:
>
>You should be able to run the proposed implementation on your computer.
>You should obtain the same results as indicated in the accompanying paper.
>These results must correspond to the original paper.

If I am made to commit potential copyright infringement to publish my replication I may have to retract my submission.

benoit-girard:
I am sorry, I am in holidays this week, thus maybe not as reactive as the situation requests it. About the equations: I am not sure it is necessary, for the current paper, to list side by side the original and the currently implemented ones. However, I do agree with @ChristophMetzner that all equations in the code (thus, the corrected ones) should be reported in the paper: this is a standard that is valid for ReScience as for other journals. 
Concerning copyright, I am talking here about equations modified from the original ones, in a completely normal scientific process, so no problem.

alexdiem:
@benoit-girard @rougier Of course I have included the equations that I have changed, but both reviewers are requesting a list of all equations in the model. Because this paper is on a reimplementation of a model (as it's supposed to be) as opposed to extending the model like it would be in other papers, copying all the equations from the model is of course a copyright issue. Otherwise I could just rewrite all papers that are behind a paywall and make them freely available for everyone. I submitted to this journal because it presented itself as a great way to disseminate your open code to other researchers with them being able to cite it if they use it. Now the reviewers make me considerably uncomfortable by demanding additions to the paper that, as far as I read it, are well beyond the scope of the [reviewer's guidelines](http://rescience.github.io/edit/). I will commit my final version tomorrow the latest and then I would like to request a final decision in accordance to the reviewers' guidelines, otherwise I would like to retract my submission and I will just leave my code for others to use with the manuscript as a readme.

alexdiem:
@ChristophMetzner Please can you post the full error message, I cannot replicate the error.

ChristophMetzner:
@akdiem I will look into the error again as soon as possible. If you cannot reproduce it, it might be possible I created it when playing with the model. Will get back to you soon.

alexdiem:
I would just like to quote the guidelines for the review process, both from [http://rescience.github.io/write/](http://rescience.github.io/write/)

> If both reviewers managed to run the code and obtain the same results as the ones advertised in the accompanying material, the submission is accepted.

and from [http://rescience.github.io/edit/](http://rescience.github.io/edit/)

> The role of the reviewer is thus to verify it the proposed submission is actually replicable. This means:
> You should be able to run the proposed implementation on your computer.
> You should obtain the same results as indicated in the accompanying paper.
> These results must correspond to the original paper.

So far I do not believe my submission has been reviewed with these guidelines in mind.

ChristophMetzner:
@akdiem Regarding the list of equations: What I requested was a full list of all equations implemented in the reimplementation. I didn't request the derivation or explanation or discussion of the equations, which are the main part of the original publication (and by restating all of that, we would run into plagiarism/copyright issues, I agree). The reason why I would like to have the equations is:
- The equations in the original paper are spread over many pages (even the Appendix)
- They also appear in the text
- They contain many errors

Therefore, a list of all correct equations (as implemented) would greatly facilitate the undertsandability of your reimplementation. Thus, making it easier to use and also increasing the chances of someone else to identify/solve the remaining erros/issues.

alexdiem:
> Therefore, a list of all correct equations (as implemented) would greatly facilitate the undertsandability of your reimplementation. Thus, making it easier to use and also increasing the chances of someone else to identify/solve the remaining erros/issues.

I do not disagree with that, but regardless I will not copy equations that I did not modify into the manuscript. This is my final decision on this matter and I would like it to be respected for the reasons I have already outlined. I will not engage in further discussion on this. I would like any further decisions to be based on the reviewer's guidelines on the journal, which I have explicitly quoted in my previous post. I will explicitly state once I have made my final changes to code and manuscript.

alexdiem:
@ChristophMetzner Regarding the code:

a)	Why do you believe that? Review the definition for the variables.
b)	What exactly are "key parameters" in your opinion and which aren't? All parameters are equally important and I have named them in accordance with the names in Witthoft2012 so that they are easily identifiable.
c)	Will explain abbreviation "eq" in the comments.
d)	Maybe, I need to review that.
e-f)	Please check that you didn't break the code yourself. I can't replicate either of the problems.

alexdiem:
@ChristophMetzner regarding the article:

b)	It is already
c)	It does now
d)	It does now
e)	Just because 0.025 is the smallest labelled tick does not mean that that is the smallest value. Please have another look at the graph.
f)	I do not see any value in adding that to the manuscript. I run the code and verify that the behaviour is the same if I really want to check.

alexdiem:
@TiinaManninen 

a) I explain it once and use the abbreviation after.
b) Can't do anything about the fact that I can only write the letter mu in math mode.
c) That's because I have to write it in math mode there and there is no way around it.
d) Have modified the text

alexdiem:
I have committed my final version and request a final decision in accordance with the reviewers' guidelines. I will only accept further requests if they concern a problem in the code only, but only if full error messages are provided and care has been taken to not have added the error oneself.

khinsen:
A quick comment on the copyright issue with respect to the equations in the original paper. Equations have no special copyright status, they are just text passages. Reproducing an equation from a paper is the same as reproducing a sentence. Reproducing many equations from a paper is then the same as reproducing many sentences. Note that this is not the same as reproducing a long text passage, as long as there are pieces in between the quoted sentences that are not taken over.

In this specific case, reproducing the equations and explaining all the quantities using different words is fair use for scholarship in my opinion.

Suggested reading: the [Wikipedia entry on Fair use](https://en.wikipedia.org/wiki/Fair_use), in particular the discussion of the four fair use factors. The issue here is the third one, "Amount and substantiality". A point to consider (which I cannot judge for lack of competence in the field) is whether the equations represent a significant part of the the contents of the original article. If the equations being discussed are, for example, those of Appendix A, the answer is no in my opinion, as an appendix can hardly be considered the core message of a paper.

rougier:
Dear @akdiem,

I'm writing here as editor-in-chief concerning your questions and the review process.

I can understand your feeling about the many modifications that are being asked but the goal of the review process is really to help the author to improve both the code and the manuscript such that others can re-use it easily. I've recently published a paper in ReScience myself (https://github.com/ReScience/ReScience-submission/pull/28) and I had to do a lot of modifications following reviewer recommendations. However, at the end of the process, I think the final paper and the code were in much better shape than the initial submission. I hope you have the same feeling about your submission. Nevertheless, we may need to make the reviewer guide more explicit and you're welcome to suggest any improvement related to your own experience.

Concerning the report of all equations in your submission, the copyright situation is not clear and it might be better not to include them as you have decided to do. In the meantime, I've written to Elsevier to know what is the legal situation because we may have the same problem in the future and we need
clear answers.

Finally, let me thank you to have made the effort of replicating a model and to have submitted to ReScience. As you may know, we are a young non-professional journal, based on a communitary and voluntarily effort and of course, we can make errors.

iampritishpatil:
This may not be the most appropriate place for me to comment, but I have been passively following the review of this paper. The points raised about including equations etc concerned me, and I tried to find any other sources. Since I found none, that shed light specifically on this kind of situation, I posted this as a question on stackexchange. Here I link to it. Perhaps someone can suggest a better place to link to it. 

https://academia.stackexchange.com/questions/94667/what-exactly-would-count-as-copyright-infringement-in-a-replication-paper

If you have any comments on this, I strongly encourage you to post them on Stackexchange, so this will be a useful reference for future, not just for ReScience, but also other replication studies in general.

TiinaManninen:
@akdiem It seems that the plot commands have been commented out. Please fix them. Manuscript still talks about overshoot of astrocytic Ca and EET. I recommend that you read your manuscript through and correct what needs to be corrected.

alexdiem:
Dear @rougier 

of course it is helpful to have many reviewers' recommendations as to how to improve the manuscript and code. However, in my particular case, I felt like the requests for changes would (a) just go on and on and on and on because every time I took great care to implement the changes, I was met with entirely new ones and (b) my manuscript was being pushed into becoming a "full" paper, rather than a paper accompanying the code I have written.

I do realise the journal is young and of course it is ok to make mistakes. Personally, for the stated purpose of this journal, which is the reproduction of previously published results to confirm the papers are reproducible, I think the reviewers' guidelines are spot on and should not be changed, but gently enforced. This journal is meant to be more about the code so I believe the paper should be as long as necessary, but as short as possible. The focus of the review should be on the code and the paper should be ensured to adequately support the code. Ensuring this should be the job of the editor. If the journal editors decide to change the reviewers' guidelines to put more emphasis on the paper rather than the code, then the journal is obviously entitled to do so, but not in the middle of a review of someone's submission. If my submission isn't treated exactly like laid out on the website, what's the point in me submitting?

With regards to my particular issue, I find it remarkable that absolutely no one appears to appreciate the dilemma I was put in regardless of whether a lawful copyright issue might arise. The fact that I replicated a paper from a particular journal should make it obvious that this journal is also within the scope for my future publications. If I republish their equations for free, regardless of whether that might actually become a lawful copyright issue or not, they might just decide to blacklist me for their journal. This cannot be in your interest for potential authors. Before I put myself into that position, I'll happily just publish not-peer-reviewed code instead of publishing here and this issue takes me to the brink of retracting my endorsement for this journal altogether.

Kind regards
Alexandra Diem

alexdiem:
@TiinaManninen I would have appreciated if you could have just mentioned where you still found mentioning of the Ca and EET overshoots. You may have experienced the phenomenon yourself, that once you have read your own writing too many times you just have trouble seeing mistakes. Hence why I removed all references to a change in Ca and EET except for one place. I have uncommented the relevant plot commands.

TiinaManninen:
@ChristophMetzner @benoit-girard

@akdiem I am sorry I was not clearer what exact sentence needed correction but thank you for fixing the sentence. I find the topic of this journal very important, thus it is important that the manuscript as well as the code are well-written and understandable. I hope that the review process has been helpful in this matter. 

It seems that the code versions online have not been updated in 7 days, thus I cannot find the uncommented plot versions. When you uncomment them and run the codes, make sure you have the last versions of the figures in your manuscript so there won't be any extra work for you later on. 

I just realized that the currents in your Figure 4 are different than what you had originally. I hope that the orders of magnitude that you give in the manuscripts are still valid, for me it looks like they are all either one or two order of magnitude different than the currents in 2013 paper but correct me if I am wrong. I am sorry, I did not realize this earlier.

benoit-girard:
@ChristophMetzner and @TiinaManninen 
Can I ask you to post a wrap-up of the status of your respective reviews: are there still points that remain to be addressed, and if yes, which ones?

ChristophMetzner:
I am satisifed with the current status of code and article, except for this one question I had asked earlier, which @akdiem wanted to explore further: 

"The original article states J_BK=I_BK/(Castr*gamma), but shouldn’t this also be negative since it is a flux of positive ions? Same for J_KIR."

The problems running the code, were introduced by myself playing around with it and I do apologize for that. I understand the concerns about including the equations and, especially since we haven't heard back from Elsevier, I am ok with them not being added.

alexdiem:
I've now had time to review the J_BK and J_KIR equations, but it doesn't appear to me that the sign of J variables has to do with the charge of the ions, but rather with the direction of the flow through the channel, i.e. out of or into the compartment.

alexdiem:
I also double checked the commits for the uncommented plot commands, and I forgot to push them. I double checked and corrected the text for figure 4.

TiinaManninen:
@ChristophMetzner @benoit-girard

@akdiem I read the manuscript and was able to plot the results presented in the manuscript. My only concern is that J_TRPV still has positive sign both in original Equation 5 that you are using and in your version of Equation 23.

alexdiem:
I just modified eq 23.

TiinaManninen:
@ChristophMetzner @benoit-girard

@akdiem Because of the change in the equation, the result plots that the new codes produce are different than earlier, but you still have the old figures in your manuscript. Mostly the results vary just a bit, but new Figure 6b plotted from the new code is very different compared to the earlier version.

alexdiem:
It seems that the best set of changes was to leave eqs 5 and 23 as in the original publication, but change the definition of J_trpv to positive sign and J_Ca negative.

TiinaManninen:
@ChristophMetzner @benoit-girard

@akdiem The signs in your version of Equation 23 in the manuscript do not match the signs in your code. Please correct in the manuscript.

alexdiem:
Thanks for carefully checking, there's always the one spot.

TiinaManninen:
@ChristophMetzner @benoit-girard

@akdiem Thanks for making the last modification! The signs were already confusing in the original manuscript, but if you plan still to make errata with the original authors, I would kindly suggest to pay attention to the signs that they are given biologically correct. I have now accepted the submission. Congratulations!

benoit-girard:
@ChristophMetzner @akdiem 
If I correctly followed the discussion, the last point standing concerns the signs of J_BK and J_KIR currents? 
@ChristophMetzner : do you still have a concern about that?

ChristophMetzner:
@benoit-girard @akdiem No, this issue has been resolved.

benoit-girard:
@ChristophMetzner are you now ready to make a final decision [accept/reject] on this paper?

ChristophMetzner:
@benoit-girard @akdiem My final decision is to accept the paper.

benoit-girard:
@akdiem Congratulations, I hereby accept the submission for publication!
@ChristophMetzner  , @TiinaManninen  thank you for your reviews!

benoit-girard:
@akdiem could you provide a list of keywords?

alexdiem:
Thank you for accepting my submission! Please use these keywords: neurovascular modelling, neurovascular unit, chemical signalling, astrocytes, neuroscience

benoit-girard:
**EDITOR**

This submission has been accepted for publication, and has been published at 
http://rescience.github.io/read/

[DOI](http://doi.org/10.5281/zenodo.890901)

rougier:
@benoit-girard Don't forget to update [Volume 3 - Issue 1](https://github.com/ReScience/ReScience/blob/master/Volume%203%20-%20Issue%201.md) file as well

benoit-girard:
@rougier I did and just submitted the corresponding PR (hopefuly).

benoit-girard:
When the last step of the "publication process" states "Close the PR without merging", it refers to the PR initiated by the author in ReScience-submission? Meaning, in this case, the PR #33 where this comment appears?

khinsen:
Exactly! Just push the "close pull request" button below.

rougier:
I cannot see your PR. Do you see it on your side ? Do you have a pointer ?

benoit-girard:
@khinsen Will close it now
@rougier Finally managed to do the PR (sorry for the delay)

PR #35, opened 2017-08-02
Review request: Detorakis

gdetor:
**AUTHOR**

Dear @ReScience/editors,

### Original article

**Title:** A Generalized Linear Integrate-and-Fire Neural Model Produces Diverse Spiking Behaviors 
**Author(s):** Stefan Mihalas and  Ernst Niebur
**Journal (or Conference):**  Neural Computation
**Year:**  2009
**DOI:**  10.1162/neco.2008.12-07-680
**PDF:**   https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2954058/
(or https://capocaccia.ethz.ch/capo/raw-attachment/wiki/2010/spinn10/Mihalas%20Niebur%20NECO.pdf)

### Replication

**Author(s)**:   Georgios Detorakis
**Repository**:  https://github.com/gdetor/ReScience-submission/tree/detorakis-2017
**PDF**:  https://github.com/gdetor/ReScience-submission/blob/detorakis-2017/article/detorakis-2017.pdf
**Keywords**:  linear leaky integrate-and-fire neuron, diverse spiking behavior, differential equations
**Language**:  Python

### Results

* [ x] Article has been fully replicated

### Potential reviewers

---

**EDITOR**

* [x] Editor acknowledgment (@otizonaizit) ``3 August 2017``
* [x] Reviewer 1 (@heplesser) ``3 August 2017``
* [x] Reviewer 2 (@pdebuyl) ``3 August 2017``
* [x] Review 1 decision [accept] ``18 August 2017``
* [x] Review 2 decision [accept] ``14 August 2017``
* [x] Editor decision [accept] ``18 August 2017``


rougier:
@gdetor Thanks for your submission. Do you have a link on the original PDF?

@otizonaizit Can you handle the submission? (I know you're busy, no problem if you can't)

gdetor:
@rougier Here is the link: http://www.mitpressjournals.org/doi/pdf/10.1162/neco.2008.12-07-680

rougier:
@gdetor Would you have a free preprint instead ?

heplesser:
@rougier I would volunteer to review.

otizonaizit:
Jep, I can do it…

otizonaizit:
@heplesser : thank you! I assigned you as reviewer.

otizonaizit:
@pdebuyl : can you review this?

pdebuyl:
I can do this next week.

gdetor:
@rougier  Try this one: https://capocaccia.ethz.ch/capo/raw-attachment/wiki/2010/spinn10/Mihalas%20Niebur%20NECO.pdf

heplesser:
@gdetor @otizonaizit I find you replication successful and your manuscript well written. I would like to request a few minor improvements to text and code before approving:

1. You mention only in the Conclusion how you inferred those parameters not explicitly given by Mihalas and Niebur. I think it would be better to provide this information already under Methods.
1. Your tables are organized slightly differently than the table in Mihalas and Niebur. It would aid the reader if you mentioned the differences briefly and stated explicitly that parameter values are identical to those in the original paper.
1. In your table 4, a parameter `t_f` appears that is not explained in the text. It appears to be the length of the trace shown in Fig 1 for each case, but needs to be explained.
1. In Fig 1, you should place the case labels A, B, C, ... outside the frame in black and ensure that there is a tick label at the right edge of the x-axis for each frame. To avoid confusing the reader, you should also point out that you use the same y-axis limits throughout, while the original figure used different limits for each case.
1. Could you add flow vectors to Fig 2?
1. I think your Fig 3 is more informative than the original one, but you should maybe point out that your figure is oriented differently from Fig 3 in the original paper.
1. You `mnn_model()` code mixes model simulation and plotting concerns. I would strongly suggest that `mnn_model()` should only simulate the model and that "prettification" is handled elsewhere.
1. External current is supplied via the `pms` dictionary as an array of current values; this array implicitly determines the duration of the simulation and assumes a certain time step. Providing simulation duration and time step as separate variables risks inconsistencies. The tidiest solution is probably to provide `I_ext` and `dt` as arguments to `mnn_model()` and infer `time`. I also do not quite understand why you build `T` incrementally as a list instead of creating it as a NumPy array right away.
1. `run_all` would be a bit more legible if you defined a function that creates each single plot and then called that function for each case, instead of repeating the same code over and over.

otizonaizit:
@pdebuyl : thank you for accepting to review! Next week is perfectly fine, take your time…

otizonaizit:
@heplesser : thank you for the thorough review!
@gdetor : you can start addressing @heplesser comments already :-)

pdebuyl:
Dear @gdetor, @otizonaizit,

The reproduction by @gdetor is well executed and runs as expected on my computer. The code
dependencies are all standard, suggesting easy re-use.

I have a list of comments that I believe should be addressed before accepting the article:

1. In the original manuscript, as far I can read, the solutions are obtained in Eq. 3.2 with
   the next firing time obtained in Eq. 3.5. I used the pdf at
   https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2954058/pdf/nihms211651.pdf , in the file
   https://capocaccia.ethz.ch/capo/raw-attachment/wiki/2010/spinn10/Mihalas%20Niebur%20NECO.pdf
   these are eqs. 4 and 7. I understand very well the rationale to choose a fixed time step
   ODE solution here but this choice and the different technique with respect to the
   original manuscript should be commented on explicitly.

2. The license for the whole repo is GPLv3. It is not suitable for the article that should
   be licensed under CC-BY.

3. The `code/README.md` file does not contain any information. It should contain a brief
   description of the Python files and the [platform
   information](http://rescience.github.io/platform/) that is requested since several months
   now by ReScience.

4. The first argument to the model integration routine is a mutable. In the case where one
   runs the code with `ipython -i` or via the `%run` IPython magic, the defaults will be
   those given by the conditional execution after the line `if __name__ == '__main__'`. This
   is a likely source of confusion and bugs for users of the code.

5. The main result of the paper is figure 1. Although the layout in a 4x5 matrix is intended
   to follow the original paper, it makes the figure's content unsuitably small. I suggest
   to change this by using a 2x10 matrix or splitting the figure, for instance.

6. A check could be added in `mnn_model` to ensure that the length of Iext is sufficient.

gdetor:
Dear @otizonaizit  @heplesser @pdebuyl ,
Thank you for your corrections/comments and suggestions. All the issues have been addressed. The reviewed manuscript and the source codes have been updated accordingly.

Thank you,
Georgios

pdebuyl:
Dear @gdetor,

Thank you for the update. You modified the article license to BSD, it really should be CC-BY :-)

    git show master:code/LICENSE.txt > code/LICENSE.txt

will give you the original one that is totally fine. BSD is of course fine for the code
directory.

The problem of the `params` mutable is still of relevance. It is now avoided in the specific
cases I mentioned in my earlier report but can still bite anyone re-using params after
importing the module `neuron_model`. A method such as "default_params" that creates and
returns a copy of a the parameters dictionary would be much safer, along with a
`params=None` default-valued argument that would call this method when it is indeed
`None`. The default-parameters-getter could also override the values with keywords arguments
to save typing later on:

    pms = neuron_model.default_parameters(a=0.005)

All my other concerns were addressed suitably.

gdetor:
Dear @pdebuyl 
Thank you for the comments. I changed the license to CC-BY (the license for the code though is now GPL2). Furthermore, I add a function that generates a default parameters dictionary and takes keyword arguments as you suggested.

article/detorakis-2017.tex
```diff
@@ -270,9 +270,10 @@ \section{Introduction}\label{introduction}
 and threshold potentials and internal currents. All the results provided in 
 \cite{mihalas:2009} have been obtained by using only two internal currents and
 thus we use the exact same number of internal currents in this work. The
-differential equations describe the subthreshold dynamics, while the 
-instantaneous threshold potential controls when the neuron firing an action 
-potential (spike) in a dynamic way. The ability of the MNN model to generate
+subthreshold dynamics are defined by a set of linear ordinary differential 
+equations, while an instantaneous threshold potential controls when the neuron
+firing an action potential (spike) in a dynamic way. The ability of the MNN
```
heplesser: Should this read "controls when the neuron fires" instead of "firing"?

article/detorakis-2017.tex
```diff
@@ -282,19 +283,35 @@ \section{Introduction}\label{introduction}
 \section{Methods}\label{methods}
 
 In order to implement the model described in \cite{mihalas:2009}, we discretized
-the dynamical system using the Euler's integration scheme. The time step is 
-fixed to $0.1\, \Rm{ms}$ for all the simulations, and the simulation time
-varies according to figure 1 of the original paper. We provide all equations
-and parameters of the model in tables as it has been suggested
-by~\cite{nordlie:2009}. 
+the dynamical system using the forward Euler's integration scheme. The time step
```
heplesser: I think it should either be "using Euler's forward integration scheme" or "using the forward Euler integration scheme".

article/detorakis-2017.tex
```diff
@@ -522,15 +534,27 @@ \section{Results}\label{results}
 All the different spiking behaviors of the model are illustrated in 
 Figure~\ref{fig:1}, where the black solid line indicates the membrane potential
 ($V(t)$), the red dashed line illustrates the instantaneous threshold potentials
-($\Theta(t)$), and the gray line shows the input to the neuron ($I_e/C$).
+($\Theta(t)$), and the gray line shows the input to the neuron ($I_e/C$). 
+The $x$-axis scale in all the panels are exactly the same as in the original
```
heplesser: "The $x$-axis scales in all panels are", not "scale".

article/detorakis-2017.tex
```diff
@@ -522,15 +534,27 @@ \section{Results}\label{results}
...
+paper (indicating the total simulation time ($t_f$), while the $y$-axis scale 
+differs from the one in the original paper. In this work the $y$-axis scale
+is the same same for all the subplots ($[-95, -25]\, \Rm{mV}$), except from
```
heplesser: "except for"

code/neuron_model.py
```diff
@@ -35,12 +46,12 @@
           'A2': 0.0,             # I₂ additive update constant
           'G': 0.05,             # Membrane potential decay term
           'C': 1.0,              # Membrane capacitance
-          'Iext': np.zeros((200,)),    # External current
           'ThetaInf': -50.0}     # Reverse threshold
 
 
 # @profile
-def mnn_model(pms=params, time=10, dt=1.0, IC=(0.01, 0.001, -70.0, -50.0)):
+def mnn_model(pms=params, Iext=np.zeros((200,)), dt=1.0,
```
heplesser: Mutables should not be used as default argument values (e.g. http://docs.python-guide.org/en/latest/writing/gotchas/). It is safer to use `None`:

```Python
def mnn_model(pms=params, Iext=None, ...):
    if Iext is None:
        Iext = np.zeros((200,))
```

code/neuron_model.py
```diff
@@ -53,70 +64,56 @@ def mnn_model(pms=params, time=10, dt=1.0, IC=(0.01, 0.001, -70.0, -50.0)):
...
-            v_plot (ndarray) : The membrane potential processed for pretty
-                               plots.
-            P (ndarray)      : The values of the instantaneous threshold
+            theta_ (ndarray)  : The values of the instantaneous threshold
                                potential at the spike-event times.
-            spikes (ndarray) : The spike trains.
+            spikes (ndarray)  : The spike trains.
     """
-    sim_time = int(time / dt)   # Total simulation time
+    sim_time = Iext.shape[0]                    # Simulation time
```
heplesser: This isn't really the simulation time, it is the number of steps. Simulation time is obtained by multiplying with `dt`. So `sim_steps` would be a better name.

heplesser (changes requested):
@gdetor Thank you for your changes. I just have a few minor language- and code-related issues left, see my inline comments.

code/neuron_model.py
gdetor: This is weird. In my last code update, the issue with the mutable has been fixed.

gdetor:
@heplesser Thank you for your corrections and comments. All the issues have beed addressed.

heplesser:
@gdetor Thank you for your updates! After your last commit f6f3165, [lines 70ff of `neuron_model.py`](https://github.com/gdetor/ReScience-submission/blob/f6f3165edf8ba6ccb1067e046e04ce34d9c2707e/code/neuron_model.py#L70) read
```python
def mnn_model(pms=None, Iext=np.zeros((200,)), dt=1.0,
              IC=(0.01, 0.001, -70.0, -50.0)):
```
so the mutable `np.zeros((200,))` is still there as default argument.
https://github.com/ReScience/ReScience-submission/pull/35/files#diff-8a081cf2e57010471dacd32d1e3805e6R70

gdetor:
@heplesser Thank you for noticing that. I updated the code accordingly.

heplesser (approved):
@gdetor Thank you for your corrections!
@otizonaizit I approve the manuscript and code for publication in ReScience.

otizonaizit:
@gdetor Congratulations, I hereby accept the submission for publication!
@heplesser , @pdebuyl thank you for your reviews!

pdebuyl:
Before publishing, I'd just like to request that the edits are made in the markdown file and not the latex file. If ReScience decides at some point to publish html versions of the articles this will be mandatory.

otizonaizit:
@pdebuyl  good catch! Thanks for pointing this out. @gdetor can you take care of this before we proceed?

gdetor:
@otizonaizit I fixed the markdown file accordingly. Let me know if you need anything more.
Thank you.

rougier:
@otizonaizit Do you need some help for the publication (I know you're busy right now)?

otizonaizit:
Thursday afternoon or Friday afternoon I'll give it a go!

otizonaizit:
@gdetor : we have problems compiling your file, the tex file is corrupted, some headers are wrong, the bibliography is included in the md file, etc... can you try fixing the issues and testing that you can generate a tex file with pandoc from the md file and then compile the tex to pdf?

otizonaizit:
Article is published with [DOI](https://doi.org/10.5281/zenodo.802285)

PR #41, opened 2017-10-26
Review Request: Mondeel, Ogundipe, Westerhoff

ThierryMondeel:
**AUTHOR**

Dear @ReScience/editors,

### Original article

**Title:** Predicting metabolic biomarkers of human inborn errors of metabolism  
**Author(s):**  T. Shlomi, M.N. Cabili, E. Ruppin
**Journal (or Conference):**  Mol. Syst. Biol
**Year:** 2009
**DOI:** 10.1038/msb.2009.22
**PDF:** http://msb.embopress.org/content/msb/5/1/263.full.pdf

### Replication

**Author(s)**:  Thierry D.G.A. Mondeel, Vivian Ogundipe, and Hans V. Westerhoff   
**Repository**:  https://github.com/ThierryMondeel/ReScience-submission-Shlomi2009/tree/Mondeel-Ogundipe-Westerhoff-2017 
**PDF**:  https://github.com/ThierryMondeel/ReScience-submission-Shlomi2009/blob/Mondeel-Ogundipe-Westerhoff-2017/article/Mondeel_Ogundipe_Westerhoff-2017.pdf
**Keywords**:  Flux variability analysis (FVA), inborn errors of metabolism (IEM), metabolic reconstruction
**Language**: Python
**Domain**:  Systems biology, computational biology

### Results

* [X] Article has been partially replicated

### Potential reviewers
**Potential editors**: Timothée Poisot, C. Titus Brown, Karthik Ram
**Potential reviewers**: Simon van Heeringen, Gökcen Eraslan, Federico Vaggi, Aaron Shifman

---

**EDITOR**

* [x] Editor acknowledgment (@ctb `Oct 31, 2017`)
* [x] Reviewer 1 (@aaronshifman `Nov 7, 2017`)
* [x] Reviewer 2 (@FedericoV `Nov 8, 2017`)
* [x] Review 1 decision accept
* [x] Review 2 decision accept
* [x] Editor decision accept (@ctb `Mar 25, 2018`)


rougier:
Thanks for your submission, an editor will be soon assigned.

rougier:
@ctb can you handle this submission ?

rougier:
@ctb 🛎 Can you  handle this submission ?

ctb:
yes.

rougier:
@ctb Thanks. Did you find reviewers?

rougier:
@ctb 🛎

ctb:
next on my list ;)

ctb:
@FedericoV would you be willing to review?

ctb:
@FedericoV bump.

@simonvh @gokceneraslan @aaronshifman interested/available?

aaronshifman:
@ctb I'b be happy to do it

FedericoV:
Hi Titus,

Metabolomics is not really my area of expertise, but I should still be able
to take a look at it.  I'm happy to help.

Federico

ctb:
thank you both!

aaronshifman:
@ThierryMondeel I've split this review into 3 independent sections for the purpose of clarity. Overall I found the code well written and presented. I had some small issues with the manuscript that should be relatively simple to address **(only after sending this I noticed that latex doesn't render in the comments, so my apologies)**

# Code presentation
The code in general is very well written and quite well commented and that is always appreciated, generally list comprehensions are less desirable than a vectorized statement - however given your use of cobra package these are unavoidable. I personally am also not a fan of the notebook format, but the fact the the critical component (the algorithm) is in a separate importable module more or less waives that concern. I do have some overall critiques of the code which should be quite simple to fix

- There are unused imports (for example `from cobra import Metabolite, Reaction`) which if they're unused should be removed for cleanliness.
- There are a few lines that are too long most of these arise from having a long comment at the end of a long print statement (see the except block from line 174 in findBiomarkers.py).
- There is a mix of tabs and spaces creating indents
- You also switch between multiline string (`'''`) and lines of comments (`#####`), stick to the multiline strings
- Lastly increasing some space between lines would significantly improve visibility
- For most, if not all of these comments a quick auto-format in most IDEs should take care of them

# Manuscript
I'm not a metabalomics person, so feel free to tell me that some of these comments are not warranted. Nevertheless, I do believe that in places things could be made clearer.

First off, the introduction is very direct, FVA is never actually defined aside from a numerical algorithm, most of your definitions in the methods section (boundary metabolites, exchange reactions, \ldots) I think could go into the introduction. This would make a much more gradual introduction as well as bringing everyone up to the same place. Also setting up the problem a bit more (giving context) could help.

I do want to highlight a concern with the manuscript which is how the replication is implemented. My first concern is that you say you replicate figures 1, and 2 in the manuscript when in reality figure 1A is replicated and figure 2 is presented in a table that doesn't quite match. For example what the original text calls ARGININEMIA you call Arginase deficiency - I imagine they're synonymous (google seems to think so), but the closer you can make your replication to the original results the easier it is to have confidence in the replication. Secondly I don't understand why this was presented as a table instead of a figure as done in the original.

With regards to the flux variability calculation section, I found it confusing in places. 

- Maximize $\pm v_i$, I assume that you just mean that $v_i$ can be positive or negative, in which case you can omit the $\pm$.
-When you introduce the concept you index by $i$, then later on you index by $k$, again I assume that these are representing the same thing - in which case keep the same indexing variable.
- $S$ is never defined in the manuscript
- With regards to $Z$, I follow your explanation to where you define $Z$ as a dot product and then I get lost, what is $c$, and what do you mean by "one or more fluxes". On the same vein, you define $Z_{max}$ in the text but $Z_{opt}$ appears in the equation - are these supposed to be that same thing.
- As one of the constraints you say that $\alpha_k \leq v_k \leq \beta_k$ where do $\alpha$ and $\beta$ come from?

I think this if these concerns were addressed that I would be better able to follow the IEM algorithm you present later.

My last concern is with your sensitivity analysis. While I applaud your going above and beyond so to speak, it seems to take up almost half of your replication. Given that this is a replication paper, I feel that the same result could have been attained with a few sentences (we tried some slightly permuted topologies \ldots different results \ldots method is only as good as the pathway annotation \ldots). Something like this would clarify that you did look into it but it's not the main focus (the replication is). 

Lastly in your references some references use the form of page numbers 1000-1020 and some use the form 1000-20. If you could please make that consistent.

# Software replication
The code does run and does seem to replicate the results found in the manuscript. Due to my not quite understanding the algorithm, I'm having difficulty picking apart the code. However given the care put into putting together the code and that it does seem to replicate the results - my guess is that I won't find anything pathological with further analysis. 

There is a part of this about the reference implementation being easy to use. I didn't have tqdm installed on any of my environments. Yes it does only take 15s to install, but it's an extra step - and at least for me it hasn't added much. Whether or not it's included doesn't sway my opinion in either direction - it's just something to consider.

ctb:
Thanks for the review @aaronshifman! @ThierryMondeel, minor suggestion, I've found `pyflakes` to be both simple and useful in identifying unused variables and imports.

rougier:
@ctb @ThierryMondeel Any progress ?

ThierryMondeel:
Apologies for the slow response. I have been on vacation for two weeks. 

@aaronshifman Thanks for the comments. I am addressing them. 

@ctb Is it preferable that I commit changes to address @aaronshifman's suggestions, or should I wait for @FedericoV to submit comments?

ThierryMondeel:
@ctb @FedericoV @rougier  I took the liberty to commit some changes to the manuscript and code in order to move this forward a bit. I hope you don't mind. 

Thanks for the review @aaronshifman! I agree with many of your comments and adjusted both the code and text as detailed below. 

# Code presentation
Following your suggestions, I removed unused imports, fixed some of the long lines and consistently applied spaces indentation. 

> You also switch between multiline string (''') and lines of comments (#####), stick to the multiline strings

I'm not quite sure what you mean here with "switching between". I use both of these but as far as I know I use the lines of comments just to clearly separate key parts of the code. 

# Manuscript
> the introduction is very direct, FVA is never actually defined aside from a numerical algorithm, most of your definitions in the methods section (boundary metabolites, exchange reactions, \ldots) I think could go into the introduction. 

I agree it probably increases the readability of the manuscriot to move the "definitions" subsection to the introduction. I did so in one of the recent commits.  

> My first concern is that you say you replicate figures 1, and 2 in the manuscript when in reality figure 1A is replicated and figure 2 is presented in a table that doesn't quite match.

I agree with your point about the naming of the metabolites and IEMs and I updated the table to reflect the original publication more accurately. Your point on Figure 1 is correct although the network indeed concerns Figure 1A the results are graphically depicted in Figure 1B (we show this in a table). But we indeed do not reproduce Figure 1C or 1D. This is now reflected in the manuscript. 

> I don't understand why this was presented as a table instead of a figure as done in the original.

With regard to your comment on the table of results that we show as opposed to a figure: this was done purely from the reproduction of the results point of view. I mean, the table shown is a direct output of the code and can (and is) reproduced by the code in the repository. Reproducing the figure seems like an unnecessary exercise since it was probably drawn in Illustrator or some other program. 

> With regards to the flux variability calculation section, I found it confusing in places.

Below I reply to each of your bullet points in sequence.

> Maximize $\pm v_i$, I assume that you just mean that $v_i$ can be positive or negative, in which case you can omit the $\pm$. When you introduce the concept you index by $i$, then later on you index by $k$, again I assume that these are representing the same thing - in which case keep the same indexing variable.

I see the confusion, let me try to clarify. v_i can indeed be positive or negative. However, I wrote the \pm as a way to indicate that the user has to choose to either maximize the positive direction or the negative direction of a flux "v_i" in flux variability analysis. This may indicate, for instance, maximizing uptake or production of a certain metabolite. If you write max. v_i, this will always give you a positive number when possible and usually not a negative one. Maximizing -v_i however, indicates maximization of the backward direction of the flux. I hope the changes in the text help clarify this.

> $S$ is never defined in the manuscript

Indeed we forgot to define S, thanks for catching that. This has been updated in the text.

> you define $Z_{max}$ in the text but $Z_{opt}$ appears in the equation - are these supposed to be that same thing.

Yes they are the same. We now consistently refer to Z_max. 

> what is $c$, and what do you mean by "one or more fluxes"

I also included more details on c, and how it can be used to maximize linear combinations of fluxes. 

> where do $\alpha$ and $\beta$ come from?

The alpha's and beta's are the bounds defined for each reaction. This is predefined in a metabolic reconstruction. I updated the text to hopefully more clearly explain this. 

> My last concern is with your sensitivity analysis.

I certainly understand your comment about the length of the sensitivity analysis section and I was expecting it to be raised. I will try to argue for maintaing it here though. 

The reason I incorporated this section in the paper is to highlight some essential things to understand about the method that are not apparent from the rest of the replication examples. I thought this would fit in the spirit of a "reference implementation" and the explanation thereof. It also serves to highlight that these perturbations of the method are possible in our Python implementation. 

However, if the pure replication is absolute essential than I have no good argument to keep it. 

> some references use the form of page numbers 1000-1020 and some use the form 1000-20.

I updated some of the reference page numbers for consistency. 

# Software replication
> There is a part of this about the reference implementation being easy to use. I didn't have tqdm installed on any of my environments. Yes it does only take 15s to install, but it's an extra step - and at least for me it hasn't added much. Whether or not it's included doesn't sway my opinion in either direction - it's just something to consider.

If it is not a major issue for the other reviewer, I would prefer keeping tqdm a part of the module. 

Thanks again for the comments, you raised many good points, let me know how you feel about the changes!

ctb:
Thanks @ThierryMondeel - @FedericoV, do you think you will you get a chance to look at this sometime soon? Appreciate it -- thanks!

FedericoV:
Hi sorry, been crazy busy.  I'll get into it this week.

FedericoV:
Overall comments:

The cobrapy library is fantastic and has a great API, but a lot of the heavy lifting for reproducing this work is done by the library itself.  As such, I would have liked to see a little more discussion about the different library functions the first time they were used.

In terms of overall reproduction: Figure 1 was done well, Figure 2 was done very well.  For Figure 1, I would have liked to see the other parts of the figure reproduced (not just the metabolites, but also the fluxes).  For Figure 2, the detective work you did to figure out what caused the minor differences between your initial simulations and the published paper was commendable.  Figure 3 is missing completely: is it particularly difficult to do?  I'll add two more comments - one about the writing, one about the notebooks.

FedericoV:
# Paper:
There are a few places that aren't very clear, for example:

### In the section defining biomarkers:
> Biomarkers are those boundary metabolites the range of exchange flux of
> which differs between the wild-type (WT) and mutant simulations.

You added a 'see below' but I see no diagram illustrating this in either the LaTeX or the PDF.

### The math notation isn't very clear.
- I follow a standard linear algebra notation (lowercase bold for vectors, uppercase bold for matrix, lowercase for scalars, etc).
- I would also use a different font for indicating mathematical symbols and chemical species.

### Misc:
I would have liked to see more discussion about the difference between $\phi = 0$ and  $\phi = 1$.  What are the implications of that?

FedericoV:
# Code:

### In Reproduce_figure1_Shlomi2009.ipynb
This:

`exchanges = [ rxn for rxn in M.reactions if rxn.products == [] ]`

Is somewhat unpythonic.  Why not: len( rxn.products) == 0?  (I hate implicit nulls but you can simplify it further).  Also a comment reminding readers that exchange reactions are pseudo reactions without reactants would help.

- In the last line, everything happens in cobrapy using the findBiomarkers function.  Reference your manuscript where you describe the process for finding biomarkers?

- The other parts of the figure are not reproduced.

### In Reproduce_figure2_Shlomi2009.ipynb:

- The work you did digging into why your work and their work differ and reconciling names was awesome.  

- Here, you have a subtly different definitions of exchange reactions than you did in Figure 1.  Exchange reactions now have no products or reactants while in Figure 1 you only looked at reactants.  Even if it is the same thing, it's a bit confusing to see them defined differently.  Further, the code and comments don't match: in the comments you say 1 and 1000, in the code it is -1 and 1000.

```
exchanges = [ rxn for rxn in M.reactions if rxn.products == [] or rxn.reactants == []]

# all inward to 1
# all outward to 1000
for rxn in exchanges:
    rxn.lower_bound = -1
for rxn in exchanges:
        rxn.upper_bound = 1000
        
M.medium;
```

- Figure 3 is missing all together.  Is that one particularly difficult to reproduce?

rougier:
@ctb Gentle reminder

ThierryMondeel:
> The cobrapy library is fantastic and has a great API, but a lot of the heavy lifting for reproducing this work is done by the library itself. As such, I would have liked to see a little more discussion about the different library functions the first time they were used.

Thanks, this is a good point. I added a short section to the methods discussing the main function from Cobrapy that is used: the flux variability analysis function. 

> In terms of overall reproduction: Figure 1 was done well, Figure 2 was done very well. For Figure 1, I would have liked to see the other parts of the figure reproduced (not just the metabolites, but also the fluxes). For Figure 2, the detective work you did to figure out what caused the minor differences between your initial simulations and the published paper was commendable. Figure 3 is missing completely: is it particularly difficult to do? I'll add two more comments - one about the writing, one about the notebooks.

Given your comments above I now included a new network drawing reproduction of Fig. 1 this time including the fluxes. I also highlighted the numerical prediction by the algorithm in this image which rendered the table I included previously obsolete. 

The reason I originally excluded Figure 3 is that the results that are discussed in it are already contained in the simulations needed to reproduce Fig. 2. However, for completeness I have now also added discussion of Fig. 3 and the image. 

>Paper:
>There are a few places that aren't very clear, for example:

>In the section defining biomarkers:
>Biomarkers are those boundary metabolites the range of exchange flux of
>which differs between the wild-type (WT) and mutant simulations.

>You added a 'see below' but I see no diagram illustrating this in either the LaTeX or the PDF.

Good catch. Adding the fluxes in the Figure 1 and 3 makes this clearer. 

>The math notation isn't very clear. I follow a standard linear algebra notation (lowercase bold for vectors, uppercase bold for matrix, lowercase for scalars, etc). I would also use a different font for indicating mathematical symbols and chemical species.

Yes that's better. I included the bm package now in the latex template and use that for the bold math notation. 
I am not sure what you have in mind for fonts to distinguish math and chemical symbols. Do you have any detailed suggestions? 

>Misc: I would have liked to see more discussion about the difference between $\phi = 0$ and $\phi = 1$. What are the implications of that?

I added a paragraph about this to the introduction. I hope that makes it clearer.


>Code:
>In Reproduce_figure1_Shlomi2009.ipynb
>This:
>
>exchanges = [ rxn for rxn in M.reactions if rxn.products == [] ]
>
>Is somewhat unpythonic. Why not: len( rxn.products) == 0? (I hate implicit nulls but you can simplify it further). Also a comment reminding readers that exchange reactions are pseudo reactions without reactants would help.

Yes good suggestion, I incorporated this change and the comment (also in the other notebooks).


>In the last line, everything happens in cobrapy using the findBiomarkers function. Reference your manuscript where you describe the process for finding biomarkers?
>
> The other parts of the figure are not reproduced.

Included a reference to the manuscript. By other parts I suppose you mean panels B-D in the original publication? In now included panel for Figure 1. C-D I excluded because the content does not have anything to do with the prediction of biomarkers directly, which is the focus of the ReScience manuscript. 

>In Reproduce_figure2_Shlomi2009.ipynb:
>The work you did digging into why your work and their work differ and reconciling names was awesome.
>
>Here, you have a subtly different definitions of exchange reactions than you did in Figure 1. Exchange reactions now have no products or reactants while in Figure 1 you only looked at reactants. Even if it is the same thing, it's a bit confusing to see them defined differently. Further, the code and comments don't match: in the comments you say 1 and 1000, in the code it is -1 and 1000.

Thanks, I made these definitions uniform now. I also improved the comments as you indicated.

aaronshifman:
Hello @ThierryMondeel thank you for the changes you made after the last round of revisions, I find that they make the paper much more understandable. I still have some minor comments

# Paper
- The word fraction is redundant

> they found a 56% fraction of the biomarkers that are correctly predicted and a fraction of 76% of predicted biomarkers that were predicted correctly

- I'm not sure I understand this - is X originating in the cell? In which case it should read "where X is understood to originate from the cell ..."
 > where X is understood to originate from or efflux to the biofluids.

- In the explanation of the flux variability calculations since you describe the size of $S$ as $m\times r$ for r reactions, could you define r earlier on as the number of reactions (and hence the number of fluxes)

- In the same part, Thank you for the clarification, I understand what you are going for with the $\pm v_i$, however that notation is quite cumbersome. The original paper uses the notation Min $v_i$ or Max $v_i$, if you wanted a less wordy statement you might use something like $v_i^\pm$ as a notation and explain it in the text.

- In the same part, I'm not sure I follow why the sparseness of S is relevant. Since you never discuss numerical methods for spareness, you can remove that sentence.

- Thank you for the clarification about $c^Tv$ term, this now makers sense.

- There is a cross-ref error after table 1 ([tbl:PKU_results])
- As an aesthetic comment when you describe the reaction for PHETHPTOX2, I assume the line breaks are because the equation doesn't fit on one like, having the $\leftrightarrow$ off to the side looks slightly odd. If you centered the arrow and wrote it as $\updownarrow$ then it would look like an equation written vertically (the same applies to the following chemical equation)

- In your replication of the original figure 1, what you call "EX_M*" is called "V*" in the original figure also in panel B the original figure is V2,V1,... whereas your implementation goes EX_M1,EX_M2,... I appreciate that given different software packages it may be difficult to replicate exactly, but if you could change at least one of those features (naming convention or order) they would be more easily comparable because at first glance the replication does not look correct due to the ordering.

- Thank you for the changes to the table representing figure 2. Now that the names match the replication is quite clearer.

- As far as figure 3 goes the replication seems quite clear, and you identified a difference between the replication and the original (the WT interval), a bit more discussion may be warranted on potential sources of differences - or if that isn't possible how much of an effect this should have on the results.

# Code
- Overall the code if clean and fairly easy to read - at times it can be a bit heavy on the list comprehensions, but that's not the end of the world. I do however have a couple of concerns about the software implementation.

- I can't see where the figures are being created, I see you have illustrator files, I can't open them right now, but would I be wrong in guessing that you are drawing the figures based on the output from the model - that's fine but should be mentioned in the article.

- For the replication of figure 3, it would be beneficial to have another cell with the extracted values that you use (EX_Met__L_e for CBS and AHCY) so that it's more clear to see so that the reader doesn't have to search through a large table. 

- Lastly the tqdm comment - it wasn't a major issue I'm fine leaving it in.

ThierryMondeel:
@aaronshifman Thanks! See below. 

# Paper
> The word fraction is redundant
>>they found a 56% fraction of the biomarkers that are correctly predicted and a fraction of 76% of predicted biomarkers that were predicted correctly

Yes, fixed. 

> I'm not sure I understand this - is X originating in the cell? In which case it should read "where X is understood to originate from the cell ..."
>> where X is understood to originate from or efflux to the biofluids.

Ok, I removed this sentence. The sentence following it should explain: "Positive flux indicates net secretion of $X$ by the cell and negative flux indicates the net uptake of $X$." 

X is is typically the extracellular form of the metabolite. Take the example of lactate: it would be produced intracellularly, be transported outward across the membrane and then efflux through the exchange reaction signifying net production. 

> In the explanation of the flux variability calculations since you describe the size of $S$ as $m\times r$ for r reactions, could you define r earlier on as the number of reactions (and hence the number of fluxes)

Ok, slightly rephrased this now. 

> In the same part, Thank you for the clarification, I understand what you are going for with the $\pm v_i$, however that notation is quite cumbersome. The original paper uses the notation Min $v_i$ or Max $v_i$, if you wanted a less wordy statement you might use something like $v_i^\pm$ as a notation and explain it in the text.

Yes ok, I see your point. I reverted to the original paper's notation. 

> In the same part, I'm not sure I follow why the sparseness of S is relevant. Since you never discuss numerical methods for spareness, you can remove that sentence.

Ok, done.

> Thank you for the clarification about $c^Tv$ term, this now makers sense.

Great! 

> There is a cross-ref error after table 1 ([tbl:PKU_results])

Good catch. Fixed.

> As an aesthetic comment when you describe the reaction for PHETHPTOX2, I assume the line breaks are because the equation doesn't fit on one like, having the $\leftrightarrow$ off to the side looks slightly odd. If you centered the arrow and wrote it as $\updownarrow$ then it would look like an equation written vertically (the same applies to the following chemical equation)

Thanks, good suggestion, included now.

> In your replication of the original figure 1, what you call "EX_M*" is called "V*" in the original figure also in panel B the original figure is V2,V1,... whereas your implementation goes EX_M1,EX_M2,... I appreciate that given different software packages it may be difficult to replicate exactly, but if you could change at least one of those features (naming convention or order) they would be more easily comparable because at first glance the replication does not look correct due to the ordering.

I agree, I fixed the ordering. I prefer to keep calling the reactions EX_M* since that is more standard in the field and more clearly emphasizes that they are exchange reactions.

> As far as figure 3 goes the replication seems quite clear, and you identified a difference between the replication and the original (the WT interval), a bit more discussion may be warranted on potential sources of differences - or if that isn't possible how much of an effect this should have on the results.

Following this comment, I tried to think on what I could do to investigate what the source of the differences might be. Recall that I used to set the influx of medium components to -1. The reasoning was just that this worked to produce the results in Figure 2. Not so for Figure 3, as you and we pointed out.

So I tried changing the influx to -10. And what do you know! The Figure 2 results are still upheld and Figure 3 now gives the same predictions. 

I adjusted the change to -10 in the text and I included the -1  vs. -10 simulations in the notebook of Figure 2 and discuss them in the text now. 

Thanks! without your "push" I wouldn't have figured this out ;)

#Code
> Overall the code if clean and fairly easy to read - at times it can be a bit heavy on the list comprehensions, but that's not the end of the world. I do however have a couple of concerns about the software implementation.
I can't see where the figures are being created, I see you have illustrator files, I can't open them right now, but would I be wrong in guessing that you are drawing the figures based on the output from the model - that's fine but should be mentioned in the article.

Yes you are correct. Figure 1 and Figure 3 in our reproduction are made in Illustrator based on the numerical results from the algorithm. I now mention this in the text. 

> For the replication of figure 3, it would be beneficial to have another cell with the extracted values that you use (EX_Met__L_e for CBS and AHCY) so that it's more clear to see so that the reader doesn't have to search through a large table.

I'm not sure what you have in mind here. The output is already in Figure 3B. The output of the algorithm is the 4 intervals depicted in black and red and I included the numerical values for the interval. No more information comes out of the algorithm. I simply mentioned the table that the notebook generates for readers that want to see exactly where the output is generated and whether I copied it correctly to the figure. 

> Lastly the tqdm comment - it wasn't a major issue I'm fine leaving it in.

Thanks!

rougier:
@ctb What's the status of the submission?

FedericoV:
I apologize for the delay in reviewing the last set of changes.

I'm completely satisfied with current state of the work - in particular, having findBiomarkers.py in the main repo makes it much easier to follow through the logic of the implementation.  I would recommend accepting the reproduction.

ctb:
thank you!

aaronshifman:
@ThierryMondeel Thank you for the changes - I have no further comments on the code or manuscript. I have no issue recommending acceptance @ctb

ctb:
I would like to look over in the next day or two - thank you very much
for your reviews!

ThierryMondeel:
@FedericoV @aaronshifman Thank you! 

@ctb Ok!

ctb:
Thanks, this is definitely an accept! Very nice discussion by all, thanks much - it made for great reading :).

I am about to go on vacation for a week, and will attend to this next weekend...

ThierryMondeel:
@ctb Great news, thanks! Enjoy your vacation :)

rougier:
@ThierryMondeel Congrats.
@ctb Just ping me if you need help during the publication process

rougier:
@ctb Are you back from vacation? Can you handle the publication? (You'll probably need help from me or @khinsen at some point).

rougier:
@ctb Can you handle the publication?

ThierryMondeel:
@ctb @rougier Anything I can do to help things along?

rougier:
I'll try to contact @ctb on twitter, he's more reactive there I think.

rougier:
@ctb Ping!

ctb:
This weekend. Apologies for the delay, everyone.

rougier:
@ctb @ThierryMondeel Re-opening the PR since the paper is not yet published.

rougier:
@ctb Can you give us some update on the publication process?

ctb:
Published as [10.5281/zenodo.1254629](https://doi.org/10.5281/zenodo.1254629)

PR #3, opened 2015-06-09
Review Request : M. Topalidou, N. Rougier

rougier:
**AUTHOR**

Dear @ReScience/editors,

I request a review for the reproduction of the following paper:
- **Interaction between cognitive and motor cortico-basal ganglia loops during decision making: a computational study**, M. Guthrie, A. Leblois, A. Garenne, and T. Boraud, Journal of Neurophysiology, 109, 2013.

I believed the original results have been faithfully reproduced as explained in the accompanying [article](https://github.com/rougier/ReScience-submission/raw/topalidou-rougier/article/article.pdf).

The  repository lives at https://github.com/rougier/ReScience-submission/tree/topalidou-rougier

```
$ git clone https://github.com/rougier/ReScience-submission.git
$ cd ReScience-submission
$ git checkout topalidou-rougier
$ cd code
```

---

**EDITOR**
- [x] Editor acknowledgment
- [x] Review 1 started (@MehdiKhamassi)
- [x] Review 2 started (@benoit-girard)
- [x] Review 1 decision [**accept**]
- [x] Review 2 decision [**accept**]
- [x] Editor decision [**accept**]

PR #43, opened 2017-11-09
Review Request: R. Shimoura et. al.

shimoura:
**AUTHOR**

Dear @ReScience/editors,

### Original article

**Title:** The cell-type specific cortical microcircuit: relating structure and activity in a full-scale spiking network model   
**Author(s):**  T.C. Potjans, M. Diesmann  
**Journal (or Conference):**  Cerebral Cortex  
**Year:** 2014  
**DOI:** 10.1093/cercor/bhs358  
**PDF:**  https://academic.oup.com/cercor/article/24/3/785/398560

### Replication

**Author(s)**:  Renan O. Shimoura, Nilton L. Kamij1, Rodrigo F.O. Pena, Vinicius L. Cordeiro, Cesar C. Ceballos, Cecilia Romaro, and Antonio C. Roque  
**Repository**:  https://github.com/shimoura/ReScience-submission/tree/ShimouraR-KamijiNL-PenaRFO-CordeiroVL-CeballosCC-RomaroC-RoqueAC-2017  
**PDF**:  https://github.com/shimoura/ReScience-submission/blob/ShimouraR-KamijiNL-PenaRFO-CordeiroVL-CeballosCC-RomaroC-RoqueAC-2017/article/ShimouraR-KamijiNL-PenaRFO-CordeiroVL-CeballosCC-RomaroC-RoqueAC-2017.pdf  
**Keywords**:  
**Language**: Python

### Results

### Potential reviewers

Potential reviewer: @mstimberg 

---
  
**EDITOR**

* [X] Editor acknowledgment (@benoit-girard  `Nov 13, 2017`)
* [X] Reviewer 1 (@appukuttan-shailesh `Nov 13, 2017`)
* [X] Reviewer 2 (@mstimberg `Nov 14, 2017`)
* [X] Review 1 decision [accept] `Apr 6, 2018`
* [X] Review 2 decision [accept] `Apr 6, 2018`
* [X] Editor decision [accept] `Apr 9, 2018`


rougier:
@shimoura Thanks for your submission. An editor will be soon assigned.

By the way, we have changed a little bit the template for submission so I may need to edit your text at the top (of the PR) to give all information. If it's ok for you, you don't need to do anything.

rougier:
@benoit-girard Can you edit this submission ? (there are some missing information at the top of the PR but I will fix it today)

shimoura:
It's ok for me. Thanks

benoit-girard:
Yes, I will edit it.

benoit-girard:
@appukuttan-shailesh and @mstimberg would you be interested in reviewing this submission?

appukuttan-shailesh:
@benoit-girard : I would be interested in reviewing this.

mstimberg:
@benoit-girard: yes, I'd be happy to review this submission, I can't work on it before next week, though.

benoit-girard:
@mstimberg confirmed to me by email that he will review this submission, but will only be able to start next week.

edit: Well, he also confirmed above, I missed that...

mstimberg:
Here's some first feedback about the code and its reproducibility only (i.e. not about details of the code, the article, or the comparison of the results to the original paper).

Everything needed to run the simulations and recreate the figures seems to be present, but it does not go as smoothly as I'd like it to, there's quite a bit of manual intervention necessary. The major stumbling block I encountered is that there's no clear instruction what *exactly* to run to reproduce the figures. The main `README.md` file states to run `python netPD.py` to "run the model for the spontaneous activity in figure 2", but this will not yet create the Figure 2. You need to run the additional script `figures/fig2.py` which confusingly produces a file `fig6.jpg` (presumably because it corresponds to Figure 6 in the original paper) in the code directory. For the other figures, the README file advises to "change the "bg_type" variable in line 41 or "stim" variable in line 45 from file netPD.py and run the simulation with the same command." but it does not state exactly which changes are necessary to these variables (it can be derived from the figure captions together with the code comments, but it should be stated explicitly). Additionally (and not stated in the README file), the user apparently needs to manually change the output file name to make the figure-producing scripts find the expected files.

As the minimal solution, I'd like to have the README file state the exact changes/commands to run to reproduce all the figures of the article. Ideally, however, the user would not have to change anything in the source code at all. For example `netPD.py` could define a function instead of being a script, so that another script could call it for all the different variants. Alternatively, it could be turned into a script that accepts command line arguments and the necessary calls to create the figures could be stated in the README file (and/or in a shell script that allows to recreate all figures in one go).

Apart from this major issue, I have a few minor comments (some of them rather suggestions):

- The `figures/fig2.py` script does not quite work correctly, the raster plot is empty due to the `xlim` focusing on the wrong time span.
- The dependencies should state the `pillow` package which is needed by matplotlib to create JPEG files. Having said that, I'd actually prefer to have an uncompressed image format in the first place (either a vector format like PDF, EPS, ..., or e.g. PNG).
- Talking about dependencies: It would be good to give an explicit way to install everything, e.g. either state a line like `pip install ...`, provide a pip requirements file or a conda environment file and the command to use it, or something else along those lines.
- Maybe mention somewhere that while you developed/tested this on Linux with Python 2 (which is of course important to state as you did),  there's no fundamental reason why this should not work on Windows/OS X with Python 2 or 3? Maybe it would even be good to have a "known-to-work" configuration for Python 3, given that Python 2 is going to be no longer maintained in the foreseeable future?
- Maybe it's worth putting a warning in the README file and/or code that the simulation takes a large amount of memory (seems to be at least 12GB on my machine)? A researcher running the code on a machine with 8GB RAM would have an unpleasant surprise...
- Another suggestion: `netPD.py` has a bit of code for making the simulation use Brian's C++ standalone mode with OpenMP for multithreading that is commented out -- maybe worth explicitly explaining (either in the code or in the README)  that it can be used by uncommenting the lines?

benoit-girard:
@appukuttan-shailesh : can you confirm that the review process is also going on, on your side?

appukuttan-shailesh:
@benoit-girard : Yes I have already started on the review. I encountered some problems similar to what @mstimberg described above, and so had some trouble reproducing the figures. So currently figuring out how to do it (rather than whether the model can actually do it - which it probably should as indicated in the manuscript).

I am away attending a workshop this week. Since the model is computationally demanding, I am unable to test it on my laptop and would need to return to my workstation. I should be able to submit my comments by early next week.

appukuttan-shailesh:
I have taken a closer look at the submission, and my overall impression is quite positive. The manuscript presents and supports the replication undertaken here quite well, and is easy to understand. It is appreciated that the authors demonstrated the need for using the exact expression (Eq. 3) for evaluating connection probabilities, and the discrepancies arising from using the approximation (Eq. 4). The source code itself is, in general, well structured and documented (BUT... see below). There are a few queries and concerns which I would like the authors to address, and I am quite optimistic that these can be resolved.

**Instructions for Replication**
-----------------------------------------------------------------
One of the ReScience requirements for a successful replication is "

> Clarity of the code and the instructions for running it. Uncommented or obfuscated code is as bad as no code at all.

The submission doesn't rate well on this front. The instructions for reproducing the results are not described well. This was my experience:
The `README.md` states:

> To run the model for the spontaneous activity in figure 2, type from the console:
`python netPD.py`
To run the experiment about dependence of spontaneous acitivity on input (figure 5), change the "bg_type" variable in line 41 or "stim" variable in line 45 from file netPD.py and run the simulation with the same command.

But this is what I actually needed to get this running:
1) `python netPD.py`
2) `mv data_raster_g4.0_w87.8.dat ./figures`
3) `cd figures`
4) `python fig2.py`
**IOError: cannot write mode RGBA as JPEG**
5) `pip install Pillow`
6) `python fig2.py`
**IOError: cannot write mode RGBA as JPEG**
[Pillow 4.3.0 not compatible]
7) `pip install Pillow==4.1.1`
6) `python fig2.py`
**fig6.jpg created**

As you can imagine, it would be really helpful for the users if the package dependencies are specified, and the exact steps for reproducing each of the figures is provided. Also, it is a bit confusing to have `fig2.py` generate a figure named `fig6.jpg` (in reference to original paper). This should be appropriately updated.

Finally `fig6` turned out to be different from what was expected:
fig6_small
As @mstimberg mentioned above, the x-axis is not properly set leading to the empty plot on the left. But that apart, the data in the other plots didn't correspond to that shown in Fig.2 of the manuscript. Am I doing something wrong?

I then changed the `bg_type` variable in line 41 in file `netPD.py` and ran the simulation with the same command as above. On running `python fig2.py` I received the following error:

> ValueError: Cannot take a larger sample than population when 'replace=False'

So in order to be able to assess the replications, I would strongly urge the authors to provide more detailed instructions. If feasible, it would be good to have a Jupyter Notebook that reproduces each of the figures so that the exact steps to be followed is available.

**Full or Partial Replication?**
-----------------------------------------------------------------
The authors have specified the replication to be complete. But I am not sure if that is the case. I believe the authors have not explored response to thalamic inputs. Accordingly, the replication does not seem to make use of the thalamic firing rate (15 Hz) specified in the original paper. Fig. 10 in the original paper discusses this.

**Other Changes Requested**
-----------------------------------------------------------------
- It would be good to have plots for Brian 2 implementation with Equation (3) also plotted on Figs. 3 and 4.
- Could you provide the histograms of excitatory and inhibitory population firing rates (similar to Fig. 7C of original article)?
- Would be useful to comment the source code with the equation numbers in the manuscript (wherever applicable).
- It would be more readable if `std_w_ex` is defined right after `w_ex` as these are STD, Mean. Currently these are in two different files (`netModels.py` and `netPD.py`). Good to have together as in original paper and as followed for Synapse parameters in `netParams.py`.

**Queries**
-----------------------------------------------------------------
- Table 1 of the manuscript mentions the initial membrane potential to be mean = -58.0 mV, SD = 10.0 mV. Is this correct? The source code seems to suggest -65.0 mV, with SD of 10% (not 10.0 mV).

- In `netPD.py`, on line 68 we have:
`neurons.v = 'v_r + 0.1*v_r*randn()'`
Is this inhomogeneity a part of the original model?

 - In `netModels.py`, on line 43 we have:
`0.3512*pA*bg_layer[r]`
What is the origin of the factor 0.3512? Is this from the original article, or derived? Would probably be useful to comment on it in the code.

benoit-girard:
@mstimberg are you about to deliver your comments on "details of the code, the article, or the comparison of the results to the original paper", in which case the authors should wait for them before answering to you and @appukuttan-shailesh ? Or, on the contrary, do you expect them to solve your first set of comments before diging into these questions?

mstimberg:
Sorry, I should have made this clearer in my reply: I'd indeed prefer if the authors sorted out the issues around running the code first before I take I another detailed look at everything.

benoit-girard:
Ok, so @shimoura: please process with the current set of comments from @mstimberg and @appukuttan-shailesh.

rougier:
@benoit-girard Any progress ?

shimoura:
Dear @mstimberg  and @appukuttan-shailesh ,

we fixed the issues you pointed and uploaded new versions of code and manuscript. Here are the specific changes:

@mstimberg :

- As suggested to make it easier for reproducibility, we changed netPD.py to a function and created netRun.py which accepts command line arguments which will produce different variants. Furthermore, we reformulated the README so that the user will have complete details on how to reproduce the figures. The reproduction of whole article is divided in different "protocols" and the user only need to state which protocol and the simulation time when executing the program (e.g. for protocol 0 and simulation time of 1 second type in the command line $ python netRun.py 0 1), see more details in the README.
- An additional change in the code was to aggregate all the scripts that produce figures from the original article in a single file named figures.py.
- All codes are inside folder (./code/). The generated figures will be created in the folder (./code/figures/) and data in the folder (./data/).
- A complete list of dependencies and how-to-install those dependencies was added in the README. Furthermore, we added warnings stating the necessary RAM, the system where the code was developed/tested and other importante information were added.
- Axis of figures and minor errors were solved.
- Although we do not use the Brian 2 OpenMP option in the simulations of the article, we provide a statement in the README where the option is explained.

@appukuttan-shailesh :

- As stated above, we tried to fix issues concerning the README where we provided more detailed instructions for reproducing the results. In addition, we also provide information about package dependencies (see answer above for @mstimberg ).
- Concerning Full or Partial Replication, please accept our apologies to this issue. At the moment of submission we did not observe the "box" and our guess is that it was automatically marked to Full. We agree to change it to Partial.  
- We solved errors on the generation of the figures and other minor erros. 
**- Other Changes Requested:** 
- We added plots for Brian 2 implementation with Equation (3) on Figs. 3 and 4, although the code does not create the plot for NEST which needs data from the original code.
- Yes, histograms of excitatory and inhibitory population firing rates were added to Fig 5 (to be compared with Fig 7 C of the original article). The code also reproduces such histograms although it is a bit time consuming if run in serial (instructions to run several simulations simultaneously are provided in the README).
- We wrote the equation references wherever applicable in the code.
- We wrote std_w_ex right after w_ex in netModels.py. As for netPD.py, the code was changed and now w_ex is an input to this function.
**- Queries**
- We solved the table problem.
- In the original NEST code we found a inhomogeneity where all the neurons start at -58 mV and distributed with a standard deviation of 10 mV. Since we didn’t find in the original article these values and the reason why this distribution we decided to adopt the neurons.v = ‘v_r + 0.1 * v_r * rand()’. Given the good agreement of the results we see that the initial values are easily forgotten after some time of simulation.
- We wrote about the origin of the factor 0.3512 in the manuscript (see DC input in page 4).

benoit-girard:
@shimoura : thank you very much for addressing the points raised by the reviewers.

@mstimberg & @appukuttan-shailesh : You may be busy at that time of the year, can you evaluate when you will be able to give some feedback about @shimoura 's set of improvements?

appukuttan-shailesh:
@benoit-girard : I am planning to go through the submission soon, and submit my feedback later today or tomorrow. I would be away from the middle of next week.

mstimberg:
@benoit-girard : I will send my feedback until the end of the coming week (before being out-of-office until January).

appukuttan-shailesh:
I appreciate the efforts taken by the authors for the revision. The instructions for replications have now been described in detail, with further features for easier replication of the figures. It is pleasing to see well-documented code!

Overall I am satisfied with the responses to my previous queries and comments. I wish to run and examine the model outputs again before submitting my decision (hope to do so early next week). In the meantime, I would like the authors to respond to the following:

**Queries**
-----------------------------------------------------------------
- Regarding inhomogeneity of membrane potential: Were there any reasons why you didn't simply adopt the same values as in the original code (mean = -58 mV, SD = 10 mV)? Were there any important differences when using this, as opposed to that in your current code (mean = -65 mV, SD = 6.5 mV)? If the initial values are found to be easily forgotten, as you have remarked, it would be best to remain consistent with values from the original code. If not, please elaborate on the discrepancies observed when using the original values in your model. I request you to add this discussion to the manuscript (see 'General Remarks' below).

- Regarding the effective factor (0.3512): Further to the description now added in the manuscript, I was curious of the origin and need for this factor. Was this employed in the original model. If so, could you direct me to the same in their manuscript. If not, could you elaborate on the need for this parameter, and the basis/origin of the equation used to evaluate it.

- Regarding partial/full replication: Is there any reason why a full replication wasn't attempted? Considering your success in replicating the other figures, it might be worth undertaking and extending this to a full replication. On the other hand, if you faced any difficulties in replicating the other figures, then this should be highlighted in the manuscript. The latter info could be highly useful to future users of your replicated model. (It is understandable if time/effort was the dictating factor. The current work should suffice for a partial replication).


**General Remarks**
-----------------------------------------------------------------
- Any additional assumptions or changes to parameters (e.g. membrane potential inhomogeneity, need for effective factor?) that were required for the replicaton should be clearly stated, preferably in a dedicated (sub-)section, in the manuscript. This aligns with ReScience guidelines for successful replications:
> '... the authors should clearly ... explain any obstacles they have encountered during the replication work'.

- Do update the submission info from a "full replication" to a "partial replication".

mstimberg:
I am afraid I have to postpone my final feedback before January. However, here are already some remarks, in case the authors intend to work on this over the holiday season :christmas_tree: :
First of all I agree with @appukuttan-shailesh that the changes in this revision greatly improved the clarity and ease of running the code. Most of my specific points have been addressed.

I do have a few more comments:

### Code instructions
* While I appreciate the installation instructions for the Anaconda environment, I feel that it has been done somewhat half-heartedly; in contrast to the instructions for `pip`, there are no *explicit* instructions how to install all packages in their respective versions. Following the instructions it is easy to end up with a situation where for example Brian has been installed for numpy 1.13 which makes installing pandas 0.19 (which has only been built for numpy 1.12) not straightforward. I'd prefer to have specific instructions (as for `pip`), or alternatively I would also be fine if the authors just stated something along the lines of "all software packages are also available in the anaconda distribution; for Brian 2 refer to the brian-team channel".
* While it is now clear and easy to run the code to reproduce most of the papers, I wonder why the authors did not apply the same approach for the comparison presented in Figures 3 and 4. I.e. instead of having to comment/uncomment lines in the script and renaming files, wouldn't it be possible to have a command line flag to chose the equation to calculate the connections, and store this information in the file name?

### Code itself
The code is well structured and commented in general (it could be nice to have the explanation of the `protocol` parameter, etc. directly in a docstring instead of in code comments, though). Reading the code there was only one thing I found confusing: there are different synapse objects that apply the weights differently depending on whether they are excitatory or inhibitory. This means that for example inhibitory weights are initialized in the same way as excitatory weights, but in their case the stored weights do not actually correspond to their effect. Wouldn't it be clearer to always have the same `I_post += w` effect but to store different weights? This would also mean that you could use a single `Synapses` object for all connections which could potentially be better for performance. But the performance is a minor point, to me clarity is more important here.

### KS test
I'm definitely not an expert for statistical tests, but I think the KS test is applied incorrectly. From the code and the article it seems that the test is applied to the empirical cumulative distributions instead of to the samples itself. I am pretty sure that this is not correct and that scipy is interpreting the provided values as samples (the KS test does look at the cumulative distribution, but this is done by scipy's test function itself). This might also explain why the authors get high p-values (at least > 0.05) for the "significant" differences.

### Article
The article is well written and presents the relevant information in a concise and clear way. There's a minor technical issue: compiling the document raises an error on my machine, due to two labels in the equation l.361-365:
```latex
\begin{equation}
\label{Eq:Ksyn_alt}
C_{i} = \left( \frac{1-T}{1+T} \right)C_{e} ,
\label{Eq:transitivity}
\end{equation}
```

shimoura:
Dear @benoit-girard : 
We would like to ask additional time so we can try to address the issues raised by the reviewers next month and improve the current submitted version.

benoit-girard:
@shimoura : sorry for my lack of recent feedback, you may take the time necessary for your improvements. If necessary, I will sollicitate you in February.

benoit-girard:
@shimoura Can you evaluate when a new version of the paper could be made available?

shimoura:
Dear @benoit-girard, sorry for the delay. We are working on the next version and we will send it at most in a week.

shimoura:
Dear @mstimberg and @appukuttan-shailesh,

we would like to thank you for the valuable remarks and constructive criticisms. We took into account the points you raised and we decided to add in this new version Fig. 10 of the original article where thalamic inputs are considered. 
To explain this new protocol we added a new topic in the section “Methods→External input→ Transient thalamic input”, and the results are discussed in the section “Results→Propagation of Transient Thalamic Inputs”. Furthermore, we included a new protocol option in the source code to reproduce this experiment which generates Figure 7 in our paper. This new protocol was also included in the README.md file.

Below you can find our answers to the questions you raised and what are the changes in the article.

@appukuttan-shailesh:

**Queries**
- We choose to rerun the simulations using the same values as in the original code for the initial membrane potential (mean = -58 mV, SD = 10 mV). All figures were replaced by new versions with this initial condition.
- We briefly explained the origin of the effective factor (0.3512 pA) in the “Methods section→External input → DC input”. 
- Our initial plan was only to replicate the network model and not the structural changes that are tested by the authors of the original article (e.g. changes in the transitivity index). In this regard, we believe that our replication was successful, specially in this new version where we included the “transient thalamic input” and again observed the same behavior as the original article. Given the remark of partial/full replication, we accept the partial but, observing that the major results allow the re-use of this model for any other purposes, we wonder if this is not enough to be accepted as a full replication?

**General Remarks**

- We added a new section entitled “Important information needed during replication”. In this section you can find the assumptions and changes in the parameters that we made to run the simulations.
- The submission is now marked as “partial replication”. Although, we would appreciate if our comments about the partial/full replication could be considered.

@mstimberg

**Code instructions**

- Following your suggestion we included in the README file the information that the “all packages are available in the anaconda distribution”.
- There is a new protocol (protocol = 4) added in the netRun.py script corresponding to figures 3 and 4.

**Code itself**

- Regarding the docstring suggestion we would like to opt to keep the current comment structure of the code.
- In the netModels.py, we merged the variables pre_eq_exc1, pre_eq_exc2 and pre_eq_inh in into a single variable pre_eq. Now the synaptic rule is the same for all connections, but the weights are stored with different values for each connection.

**KS test**

- As you mentioned, we were using the wrong input values for the scipy function. Thus, we opt to implement the KS test ourselves and use the KS-statistics instead of the p-value. The results remained consistent (see in the pdf).

**Article**

- We fixed the label problem.

benoit-girard:
@mstimberg @appukuttan-shailesh : the authors have proposed some feedback on your remarks almost two weeks ago. Do you think you can provide your comments on this new version of the paper soon? Is the 1st of March an acceptable deadline for you? (sooner would be even better, of course).

appukuttan-shailesh:
Yes that should be fine. I shall go through the changes and respond early next week.

mstimberg:
@benoit-girard Apologies for the delay, vacations got into the way and I also underestimated the time it takes for the full set of simulations to run through. I will run the last simulations over the weekend and will submit my review next week.

appukuttan-shailesh:
I am happy to see that a new section titled "Important information needed during replication" has been added to the manuscript. This significantly clarifies the model replication process, highlighting the information missing (or hidden) in the original article.

Regarding the extent of replication: I agree that the present work replicates the main results of the original article. The [guidelines for reviewers](https://github.com/ReScience/ReScience/blob/master/reviewer-guidelines.md) states that: "A full replication covers **all** the results of the original work". I am not sure how strictly this requires to be adhered to. All figures pertaining to results, excluding Fig. 9 of the original article (dealing with target specificity of projections), in the original article seem to have been satisfactorily replicated (pending verification by re-running simulations; see below). I would like the editor (@benoit-girard ) to take a call on this, or advice on the flexibility in guidelines.

I am still in the process of running the simulations (long runtimes!). Some of them failed. E.g.
`python netRun.py 2 5.0` gives:

> _exec_command_posix failed (status=-1)
> [Errno 12] Cannot allocate memory: /bin/bash
> 
after running for a couple of hours, and 
`python netRun.py 4 60.0` gives:
> ERROR      Brian 2 encountered an unexpected error.
> ...
> MemoryError

The PC has 32 GB of RAM, so I believe that shouldn't be the problem. I will be re-running these to check if they persist. It would probably take me a few more days to run the simulations, as I can't do this in parallel while I am working (too resource intensive).


### Minor Changes ###
1) netModels.py, line 69: I believe there is no scenario where stim == 2 (stim = 0 or 1 in netRun.py). This check could be removed.
2) netModels.py, line 134: same as above
3) README.md, line 201: (changing variable **serial**) Change to line #83 (not #74)
4) README.md, line 202: (changing # of cores) Change to line #84 (not #75)

mstimberg:
Again, apologies for the delay, running the simulations takes indeed much longer than I expected. Similar to @appukuttan-shailesh, I had major problems running the simulations for Fig.5c and Fig.6 (in my case on a machine with only 16GB of RAM). After running a few simulations it always crashed with a lack of memory (weirdly during the compilation of code, apparently). I don't see anything wrong with the authors' code, if anything this might be a Brian issue. I could potentially run it on a machine with more RAM, but this might take me a while to organize and setup and I'd prefer not to delay the review any longer. What I did for now is to simply add a little code snippet right at the start of the `runParams` function in `netPD.py` so that it will not re-run parameters it already simulated:
```Python
    if filename is not None and os.path.exists(filename):
        print('%s already simulated.' % filename)
        return
```
Then, I use a brute-force approach to re-run the `netRun.py` script for each figure until it worked...

Also, I had to set `serial=True` in `netRun.py` -- the README and comment indicate that this was supposed to be the default value. Maybe `serial=False` was also the reason why @appukuttan-shailesh's simulation crashed (running 8 simulations in parallel needs about 100GB of RAM, I think)?

I'm still waiting for the results to replicate Fig. 6, but... they will take a while. Each run for a single parameter combination seems to take between 40min and 60min on my machine, and given that there are 450 combinations that are explored I am expecting a total runtime of about... 2 weeks! Given that I cannot run these simulations in parallel on my machine, I don't see any way to bring this time down in a significant way. Would it maybe be sufficient to run the simulations for a shorter time, e.g. for 1s instead of 10s? Do the authors know for what "biological time" the authors of the original study ran the simulations? I can't find this information in their paper

Apart from this figure, I can confirm that the provided code generates the figures presented in the paper (I did not run the original NEST code for the orange curves in Fig. 3/4, though). I do get tiny quantitative differences in the histograms of Fig. 5C, but nothing that affects the results in any meaningful way.

I also think that the results qualitatively match the results of the original authors (with some small quantitative differences) and therefore agree with the authors' conclusion that they successfully replicated the original work. From my side, I'd be happy to call it a "full replication", but I agree that the guidelines are not very clear about this.

I thank the authors for adressing my concerns concerning the code (e.g. by adding the new protocol) and I'm happy with its current state (apart from the minor `serial=False` mistake). I also appreciate the new "Important Information" section in the article, it is very helpful to have details like the wrong thalamic input rate spelled out explicitly.

I have a few remaining issues with the text/README, but most of them are pretty minor:
* The installation instructions do not advise to install weave (for Python 2) or Cython (Python 2/3) -- this means that simulations will use the fallback numpy mode which will probably slow simulations down even further (for users that do not activate the standalone mode manually).
* In my opinion, the description of the application of the KS test could still be improved. I don't find the last sentence of Methods->Measures ("The null hypothesis that two sample sets were drawn
from the same distribution is accepted, for an error α = 0.01, if the KS-statistic ≤ 1.63.") very helpful (no need to state any explicit values here, this will be done later) and it is phrased a bit sloppily (I would not say that a null hypothesis can be "accepted"). I'd personally opt for simply removing it. Then, the results section just above Fig.3 only states a p value instead of the KS statistics, while later both KS statistic and p value are mentioned (but with α=0.01 which seems unnecessarily strict). However, when you *have* significant differences, you do not state neither p value nor KS statistics. I'd also suggest rewriting the paragraph on page 8 starting with "When comparing the network constructed in Brian 2" to make it clearer what you found (no difference between NEST and Brian with correct equation, some differences to Brian with "incorrect" equation). Currently the flow is a bit confusing (first sentence states "all layer-specific comparisons... indicate no significant differences", second sentence starts with "The major discrepancy was found for L5e").
* In Results->Propagation of Transient Thalamic Inputs:
    * I find the sentence "By comparing histograms in Fig. 7B with the ones in Fig. 10 B of the original article one can see the similarities." not very helpful. It should either describe the qualitative properties of the histograms and state that they match to Fig. 10B of the original article, or simply state that you think that it replicates the results from Fig. 10B without going into further details, without referring to rather vague "similarities". 
    * "Brian2" --> "Brian 2"
* The formatting of subplot references (e.g. "Fig. 5C") is not entirely consistent throughout the text (space between number and letter or not, letter in bold or not)

benoit-girard:
Regarding the Full vs. Partial replication level, I would tend to think that if one figure hasn't been reproduced, it's partial. However, in previous ReScience publications, I think I remember we haven't been that strict so I have a doubt...
@rougier : What do you think about it? And also: how important is that? I mean, does partial vs. full clearly appear in the finale version of the paper?

rougier:
There's no real definition, I only added it when I created the template. I should have think twice...
In my mind, a partial replication meant some results we replicated but some others were not replicated because the author cannot make it run. In the end, we should have 4 categories:

* **Full replication (++)**: Success in replicating all results
* **Partial replication (+)**: Success in replicating  some results
* **Partial replication (+-)**: Success in replicating some results, Failure in replicating some others
* **Failed replication (--)**: Failure in replicating any results

khinsen:
Four categories are not quite enough, because "failure" can mean two things:
 - tried to replicate, should have worked, did my best, but the results don't agree
 - insufficient information in the original work, so it's not even worth trying

I can't see from a quick look at the review process why Fig. 9 was not reproduced. @appukuttan-shailesh, can you summarize the reason(s)?

appukuttan-shailesh:
@khinsen : From my understanding, the authors haven't attempted to replicate Fig. 9, as their focus was on replicating the other results. The authors have said:

> Our initial plan was only to replicate the network model and not the structural changes that are tested by the authors of the original article (e.g. changes in the transitivity index). In this regard, we believe that our replication was successful...

Considering their success in satisfactorily reproducing the other results, I expect the remaining result(s) to also be replicable via their model. I imagine time/effort required was probably partly an influencing factor (especially in view of the fact that the simulations involved in this study are really long). As @mstimberg  has commented, the major results have been replicated, and therefore I too would be happy to consider it as a "full replication".

appukuttan-shailesh (changes requested):
Also on line #134 of netModels.py

shimoura:
Dear @appukuttan-shailesh and @mstimberg:
 
thank you for the comments. We uploaded a new version of code, readme and manuscript with your minor suggestions fixed.
 
Concerning the memory problem in the sweep simulations:
Unfortunately the code (and the original NEST) is indeed very demanding, as @mstimberg said there are 450 combinations to be explored in the already large network by itself. For the specific case of Fig 6, we used parallel simulations (with 8 cores and 128 GB RAM), although we provided the option to run it in serial. Running this number of simulations in serial is possible but should take quite long time and, as you noticed, accumulation of memory. One way we tried to avoid memory accumulation was to use a garbage collector function, such as gc.collect(). It helps, but still one would need quite a lot of memory (which we are not sure where it comes from) if the parallel option was not used.
 
One possible solution for verification of this figure could be to check a smaller portion of the diagram. In the case this suffices, we uploaded a new script (fig6_sample.py) only to check if the selected smaller portion of the diagram agrees with our figure. To use this script, you would do the following:
in netRun.py change lines 145 and 146 the intervals of g_values and bg_values to a smaller portion, for example: 
    g_values = np.arange(2.0, 4.5, 0.5);  # relative inh. synaptic strength values
    bg_values = np.arange(3.0, 6.5, 0.5);  # background rate values
Then, in fig6_sample.py, change the lines 18 and 19 to the very same interval you chose above. 
First run the code with the protocol of Fig 6, then run fig6_sample.py. 
 
This above example, delivers the following figure to us:
 
fig6_sample

An additional possibility, which could be combined with running the smaller portion, is running the code for 5s instead of 10s which would reduce quite a lot the computations. The original model actually uses 5s, we only used 10s to get better statistics. For a shorter time, change the protocol option to
 
python netRun.py 3 5.0
 
Also, there is no mention in Fig 6 about the number of points considered, we used this number because it gives the same results but one could use a smaller number of points.

appukuttan-shailesh:
Just to give you an update: I have one last simulation (protocol 3) pending, and its already been running for around a week (therefore deciding to let it proceed to completion). @mstimberg : I was wondering if we had the possibility of running this on the NSG? It appears to support Brian; not sure about Brian2.

mstimberg:
I'm also still running the simulations, but now with a 5s instead of 10s runtime as suggested by @shimoura. I should have the results by Monday.
@appukuttan-shailesh : Running things on NSG would have been a good option, but unfortunately they only support Brian 1 for now. I will  contact them about this, but it will be too late for this review, I think.

mstimberg:
To give another update:
 I'm happy with the authors' changes to the text and README files, they have addressed my concerns with one exception: I still find the paragraph in Results->Spontaneous Activity starting with "When comparing the network constructed in Brian 2..." hard to parse. The first sentence seems to say that Brian with equation (5) does not show a significant difference even in layer 5e and 23e for firing rates compared to NEST/Brian with equation (3) which seems odd given the pronounced difference in the distributions. Then I don't quite understand what the second sentence refers to with "when comparing the mean firing rates and their standard deviations, _in general_, ..."

I run the simulations to replicate Figure 6, but with 5s instead of 10s. The plots for fixed inhibitory strength/background firing rate look very similar, but the AIness colour plot shows pronounced differences:
fig6

I'm still running the reduced sample with the full 10s as suggested by the authors.

shimoura:
Dear @mstimberg

We revised the method used to calculate the KS-statistics and we decided to change how to estimate the accumulated histogram. To make it simpler we changed the ‘doane’ method to ‘sqrt’ (number of bins is the square root of data size) of the ‘histogram’ function from numpy package. The ‘doane’ method was tested before for the first version of the code that we submitted, however, as you remember, we were using the scipy function in the wrong way.
Now we tested again and we observed that changing to the ‘sqrt’ method, the results make more sense with what we can see on the accumulated histograms (Figures 3 and 4). So, the discrepancies that we commented on the manuscript now correspond with the measures that didn’t pass in the KS test.

About the Figure 6, the result for 5s may vary for each simulation, mainly for high values of g with low background rate. For these cases, it might happen that some layers fire at very low firing rate (less than 1Hz), which will influence strongly in the CV measure. The CV measure is based on the ISIs, thus we need a certain amount of spikes to start to compute this statistics efficiently. The figure below exemplify this situation using g = 10.0, bgrate = 6.0 Hz, where we can observe CVs<0.7 if we run one simulation for 5s and all CVs>0.7 if we run for 10s. 

fig

To run for 5s, the ideal is to run multiple simulations for each parameter combination and take an average. Although the authors of the original article didn’t mention if they take an average, we believe that this was the case. The other possible solution is to run for a longer time, which was our approach and the reason we decided to run for 10s.

mstimberg:
@shimoura , thank your for the explications and the update. From my side, the explication of the observed differences with the references to the KS test are now fine.

About Figure 6, I see your point and I can reproduce your result that you show in the figure. However, this leaves me in the position that I cannot reproduce your results without running it for the full 10s as you did (or running it multiple times for 5s), which would take me a significant amount of time to do (I could try to run it on my university's cluster, but I never used it so far...). I'm also in contact with the NSG maintainers about running Brian 2 simulations on their cluster -- in contrast what I said earlier, one *can* already run Brian 2 simulations, albeit it is not very useful at the moment because one cannot run simulations in parallel.  If @appukuttan-shailesh managed to reproduce the authors' results with full 10s run, I'd be willing to take the reproducibility as confirmed. But then, it might be against ReScience's policy if only one of the reviewers completely replicates the results (CC: @benoit-girard) ? While looking at the code I also realized that "protocol 3" (i.e. the simulations for Fig. 6) does not explicitly set the seeds (in contrast to protocol 2). It might therefore be -- especially if the statistics are not very robust -- that even a simulation with 10s per run might not give the exact same results.

appukuttan-shailesh:
As it turns out my simulations ended yesterday.... right at the very end it _crashed_ for some inexplicable reason. But fortunately all the individual simulation outputs are locally saved. Using those I was able to generate Fig. 6. I should be able to assess the figures early next week and submit my responses.

appukuttan-shailesh:
@mstimberg : find below Fig. 6 that I was able to produce by running the original protocol for 10 s in serial (in case you wish to use the same to draw any conclusions)
image

(more soon)

appukuttan-shailesh:
Finally getting back to this review. I ran the simulations in serial model (thanks to @mstimberg for alerting me that the default value of `serial` wasn't as expected). Protocol #3 (for producing Fig. 6) is the longest and took well over two weeks! So (as a warning to future users) this should probably only be run on clusters or supercomputers, and not on local workstations.

Overall, I am happy with the replication of the figures using the Brian 2 variant of the original model. As @mstimberg mentioned, there are minor differences in some of the figures (such as Fig. 5C, 6, 7). Some of this, I believe, arises from the inherent stochasticity of the model and could be resolved by managing the seeds. The differences anyhow are minor, and do not affect the interpretation.

As previously conveyed, I am pleased to see the addition of the new section titled "Important information needed during replication" with info that is vital for a successful replication (highlighting omissions and mistakes in the original article). I too wasn't very satisfied with the final statement of the section titled "Propagation of Transient Thalamic Inputs"... "_one can see the similarities_" . This has now been updated, and though better than before, could probably still be reworded (I personally find the term '_think_' not too confidence inspiring!). But this is a very minor remark, which the authors could take into consideration.

The authors opted not to attempt replication of one of the figures (Fig. 9 of the original article). Unless this contradicts ReScience policy, I am happy to accept the current submission as a "full replication" as in my view the major results have been replicated satisfactorily. The earlier discussion about full/partial replication got suspended without us arriving at a decision/conclusion, and probably needs more clarity going forward.

mstimberg:
To give a (potentially final) summary of my review:
I am happy with the text and code, in my opinion this article can be accepted as it is for publication in ReScience. It replicates the results in the original publication well, and as @appukuttan-shailesh I would accept it as "full replication", but I don't have a strong opinion on this, either (does it actually get mentioned anywhere in the final article?) despite the missing Figure 9.
I was able to reproduce all figures except for Figure 6 with very minor differences and I can therefore confirm that the provided code produces the figures in the article. For Figure 6, the result provided by @appukuttan-shailesh also matches the figure provided by the authors enough to not make me doubt their results. However, if @benoit-girard thinks that it is necessary for me to replicate this result myself, then I will need a couple of additional weeks to get around doing that.

[A final remark not as a reviewer but as a Brian 2 developer: it might well be that the memory problems we encountered are due to a bug/regression in Brian: brian-team/brian2#933 ...]

benoit-girard:
From what I read here, it seems that the paper is about to be accepted.
@rougier @khinsen : what do we do about this "full replication thing"? I am in favor of "full replication" (but it does not seem to be very important). @rougier suggested adding categories (but can we really define a full set of categories covering all the possible configurations?). Where do we go with regards to this? Do we stick to two? Do we label this paper "partial" or "full" replication?

khinsen:
I vote for a pragmatic approach. Our categories aren't clear and haven't been applied consistently in the past. We should apply the rules valid at submission time for each paper, which means "not quite clear" for this one. Since the reviewers are happy to call it a full replication, so be it.

This doesn't mean we shouldn't clean up our categories, but that's a different issue (soon also in the GitHub sense of the word).

mstimberg:
Just as a minor comment from my point of view about the "full replication/partial replication" issue: I think the question is not extremely pressing as this category does not appear anywhere in the PDF or the Zenodo archive, AFAICT. Of course, part of the appeal of ReScience is that the reviews are public as well, but I'd be certainly more cautious about calling the replication "full" if this was a big visible label on the paper.

benoit-girard:
@mstimberg @appukuttan-shailesh 
Dear reviewers, are you satisfied with the current state of the paper, or do you require additional changes (even minor) before acceptance?

mstimberg:
@benoit-girard Yes, I am satisfied with the current state of the paper and recommend its acceptance (noting that I did not personally replicate the simulations for Figure 6, as explained in my earlier comments).

appukuttan-shailesh:
@benoit-girard : Yes, I am happy with the (current) revised submission and recommend its acceptance.

shimoura:
@mstimberg @appukuttan-shailesh  Is it ok to delete the script fig6_sample.py? This script was used just to generate part of the figure 6 as commented on March 2, which is not part of the submitted article.

mstimberg:
@shimoura : From my side, yes, I think it would be better to delete the script to avoid confusion.

appukuttan-shailesh:
@shimoura : Yes please go ahead.

benoit-girard:
OK, so the review process is over, both reviewer accepted the paper. 
I will carry out the (relatively complex) publishing process in the coming days.

rougier:
Yes, sorry for the complex publishing process. We need to write a (Python) script to try to handle it automatically.

rougier:
@benoit-girard Any progress with the publication?

rougier:
@benoit-girard 🛎

benoit-girard:
I'm on it today...

benoit-girard:
@shimoura sorry for the long publication delay. Can you provide me with a set of keywords?

shimoura:
Dear @benoit, here are the keywords: 
cerebral cortex; cortical microcircuit; large-scale models; layered network; specificity of connections;

benoit-girard:
Thanks !

benoit-girard:
**EDITOR**

This submission has been accepted for publication, and has been published at https://github.com/ReScience/ReScience/wiki/Current-Issue

[DOI](https://doi.org/10.5281/zenodo.1243268)

benoit-girard:
The version 1.1 of the paper has a correctly compiled bibliography. Its new DOI is:


[DOI](https://doi.org/10.5281/zenodo.1244116)

PR #46, opened 2018-02-07
Review Request: Senden Schuecker Hahne Diesmann Goebel

MSenden:
**AUTHOR**
Mario Senden; Jannis Schuecker; Jan Hahne; Markus Diesmann; Rainer Goebel

Dear @ReScience/editors,

### Original article

**Title:**  A Neural Model of the Saccade Generator in the Reticular Formation.
**Author(s):**  Gancarz, G. & Grossberg, S
**Journal (or Conference):**  Neural Networks
**Year:**  1998
**DOI:**  10.1016/S0893-6080(98)00096-3
**PDF:**   [Gancarz-1998.pdf](https://open.bu.edu/bitstream/handle/2144/2348/98.014.pdf?sequence=1)

### Replication

**Author(s)**:   Senden, M., Schuecker, J., Hahne, J., Diesmann, M. and Goebel, R.
**Repository**:  [Senden-2018](https://github.com/MSenden/ReScience-submission/tree/senden-schuecker-hahne-diesmann-goebel-2018)
**PDF**:  [Senden-2018](https://github.com/MSenden/ReScience-submission/blob/senden-schuecker-hahne-diesmann-goebel-2018/article/senden-schuecker-hahne-diesmann-goebel-2018.pdf)
**Keywords**:  saccade generation; rate neurons; reticular formation
**Language**:  Python; NEST
**Domain**:  Neuroscience

### Results

### Potential reviewers
---

**EDITOR**

* [x] Editor acknowledgment (@otizonaizit) `13 February 2018`
* [x] Reviewer 1 (@apdavison) `13 February 2018`
* [x] Reviewer 2 (@gdetor) `16 February 2018`
* [x] Review 1 decision [**accept**] `23 April 2018`
* [x] Review 2 decision [**accept**] `23 April 2018`
* [x] Editor decision [**accept**] `4 May 2018`


rougier:
Thanks for your submission. An editor will be assigned soon.

rougier:
@otizonaizit Can you edit this submission ?

rougier:
@otizonaizit Can you edit this submission ?

otizonaizit:
sure!

otizonaizit:
@heplesser : could you review this submission?

otizonaizit:
@apdavison : could you review this submission?

apdavison:
Possible conflict of interest: I'm a co-PI with Markus Diesmann on a grant (Human Brain Project)

otizonaizit:
> Possible conflict of interest: I'm a co-PI with Markus Diesmann on a grant (Human Brain Project)

@apdavison : if you were not involved in the work leading to the present submission this is fine, in my opinion, but let's see what @rougier thinks about it...

otizonaizit:
by the way, @heplesser may be in a similar conflict, but again, if @rougier agrees with me, I see no conflict of interest if @heplesser was not involved in the work leading to the present submission

apdavison:
I confirm I was not involved in the work leading to the present submission

rougier:
Thanks @otizonaizit. 
For me it's fine. And the review is public such that anyone thinking the review is not fair can raise a hand.

apdavison:
In that case, I'd be happy to review this.

heplesser:
@otizonaizit I am collaborating extremely closely with Markus Diesmann, so I am afraid I have too much of a conflict of interest. I have also been the main reviewer of the NEST extensions required to make this paper possible. Therefore, I have to decline.

otizonaizit:
@heplesser : OK, thank you for letting us know.

otizonaizit:
@rossant : could you review this submission?

rossant:
@otizonaizit sorry, no, I am not available

otizonaizit:
@rossant : Ok, thank you for letting me know.

otizonaizit:
@gdetor : could you review this submission?

otizonaizit:
@rougier, @khinsen : the code in this submission is setting up a model and running simulations using the [NEST Simulator](http://www.nest-simulator.org/). The simulator is not included in the submission, of course.

I am not sure what is the policy about this kind of replications, where the code in the submission is mostly about plotting and setup, whereas the whole logic is actually hidden behind a big external library.

My impression is that, for the work to be considered a replication of the original work in the ReScience sense, and for the results in this submission to be reproducible for future readers in the ReScience sense, external dependencies should be, at least for Python code, restricted to the "standard" scientific Python stack, i.e. `numpy`, `scipy`, `matplotlib`, maybe `pandas`, and nowadays maybe `scikit-learn` and `scikit-image`? It is my very specific and humble opinion, so I wonder what @rougier and @khinsen think about it?

On the other hand, by quickly skimming through the original paper (58 pages!!!) I couldn't find a specific reference to the software that was used to simulate the model back in 1998. So, in this respect this submission definitely increases the replicability of the original paper.

khinsen:
I don't know much about NEST, nor about its domain of application, so my point of view will have to remain rather abstract.

Pretty much all studies rely on some packaged software, so obviously we must allow that. As you say, the "standard" SciPy stack is... standard. How much else is standard is a domain-specific question and can only be answered by the reviewers and editors of a specific article. I don't think we should have a finite list of acceptable libraries, because that would exclude replications from many fields.

So, if NEST is "standard" in its field, then there is no problem from my point of view.

The trickier question is what to do with less standard code dependencies. Not allowing them might again come down to excluding a lot of topics from ReScience. Allowing anything at all opens the door to obscurity. A Python script that merely calls a function from a library written in obfuscated C is not acceptable, even if the obfuscated C code is open source. But where to draw the borderline?

I am tempted once more to leave the decision to our reviewers. The question that reviewers should ask themselves is: Given what you know, or what you can learn with reasonable effort, about a software dependency, do you feel that you understand what it does well enough to judge its adequacy for the question being studied? In case of doubt, write a sentence about it in your review. And if you have the impression that a dependency is inappropriate, or could easily be replaced by five lines of code, you should complain about this in your review as well.

otizonaizit:
About NEST being standard or not: There are several neuronal simulators, all of them with some overlapping features and some unique features. There are heated discussions in the field which simulator _should_ be standard, but I don't think that a consensus has been reached and I would name at least three simulators among the big ones: Neuron, NEST, and Brian.

In this submissions is definitely not feasible to replace NEST with  5 lines of code, not even with 1000 lines of code, and this is exactly the point that makes it questionable for me.

To understand the implementation of the models and the results of the simulations, you need to understand NEST, which is a huge bundle of C++ code (~100,000 LOC) with some Python (~16,000 LOC). And you can't even argue that any neuroscientist dealing with neuronal models knows NEST, because many of them are using other software to run their simulations. To me it seems a bit odd to leave the decision to our reviewers alone. I'd like to give them at least a direction or an indication in terms of general principles. I tried to let this paper be reviewed by one of the NEST developers, and he had to refuse because he reviewed the code that went into NEST which allowed this submission to be written. So, you see? That reviewer would understand exactly what the whole thing is about, but I am not sure this is a good criterion in general...

On the other hand, it is important to note that while I think that the original paper _can_ be replicated by writing some understandable Python code from scratch, that code wouldn't be usable in practice to do proper science with it because of efficiency issues, which is the reason why big neuronal simulators have been developed...

iampritishpatil:
Just chiming in as someone who has used the said simulators before, there is often an advantage to use a standard simulator. It is that it is much less likely to contain mistakes, because such simulators are written by many people together, and have a lot of users, so bug reports go in. This becomes more and more important as the complexity of the simulation increases. 

So, I'd think it is better that the simulation is written in a very abstract form using a standard simulator, than to use a custom written python code. 

Also the other advantage is that its very easy to extend simulations written in a simulator, where you can just swap some objects for another, or add another layer with just a few lines. So in terms of use fullness, its way more useful.

So as long as the simulator library code is open access, in principle, it is possible to go as deep as one wished in understanding the code, and as such is really an open replication.

khinsen:
@otizonaizit, @iampritishpatil: Thanks for the additional explanations! I think that now I understand the situation very well, because it is extremely similar to the one in my own field of research, biomolecular simulation. The big difference is that in my field, no one does replications, so ReScience doesn't have to worry about it ;-)

I have actually spent quite some time thinking about this kind of problem, and I am working on a paper proposing a possible solution in the not-so-near future (ten years would be an optimistic estimate). This also means that I don't believe there is any good solution right now.

@iampritishpatil: Your arguments are all very valid from a software engineering point of view, and also from a research productivity point of view. But that's not what ReScience is about. Replication work is done to establish the scientific validity of a computation. It's *validation*, whereas software engineering is about *verification* of formal correctness criteria using tests etc. 

Given that there is no good solution, what we can think about is the best possible one for now. The goal of ReScience is to improve quality standards in computational science. Which line of action is best suited to reach this goal? I'll think about it and come back. In the meantime, I'd be happy to hear other people's ideas on this.

damiendr:
For the simplest types of networks (point neurons, no delays, euler integration) it is possible to write a minimal simulator in about 200 lines of code. In pure Python speed would be a problem of course.

The issue of divergences between different simulators is something that has come up quite a few times in the comp neuro community. I remember a slide showing how the same model behaved differently in two simulators — although both provided an apparently identical "LIF" neuron, they made different implementation decisions such as whether to broadcast spikes at the start or at the end of a timestep, etc. Unfortunately I can't find the article where that came from, I'll keep looking...

gdetor:
@otizonaizit Yes, I can handle the review of this submission.

apdavison:
I don't really understand the problem that @otizonaizit is seeing here. NEST is an external package, but so are `scikit-learn` and `numpy`. NEST is open-source, on Github, has possibly been around for longer than `numpy` (when I started using NEST, I was still using `numarray` and `numeric`), and has extensive unit and integration tests, so I don't see how it is more opaque than `numpy`.

As a general principle, I don't think it matters whether an external package or tool is "standard". If it is open source and can be understood and installed by the reviewers, then I think it is a valid basis for a replication.

apdavison:
Having reflected on this some more, I think it is _preferable_, for both reproducibility and reuse, for the heart of a replication to be in an external, well-tested, continuously maintained package. This greatly reduces the risk of bit-rot, and may make the model more widely and easily available to others if the external package is a popular one.

pdebuyl:
When authors submit a replication that relies on, say, Python, NumPy, and scikit-learn, we can reasonably suppose that someone willing to replicate the ReScience paper in 10 years will find these dependencies. The Python Package Index stores releases of software. Linux distributions store several releases of Python, of NumPy and of scikit-learn. scikit-learn is also on Zenodo.

Maybe the question is not that much how *standard* the software is but how likely it is to remain available (both being often correlated, should I guess). We cannot foresee the specific situations, however :(

So I though I'd look up NEST and they are on Zenodo :-) https://zenodo.org/record/882971

Having a detailed list of the dependencies and verifying that they are reasonably well archived seems the current state of the art in computational replications and I believe that ReScience helps to establish this state of the art. Improving the state of the art is also important, also possible within ReScience, but is kind of a bonus right now.

khinsen:
Here's my second round of thoughts. The starting point: what are the desirable characteristics of replications published in ReScience?

1. Independence. The software being run should be as different as possible from the one used in the original paper. Running the same software again is not a replication, but verification of reproducibility. Running a different script that calls the same library with parameters that the replication author has deduced is better, but still far from an independent replication. A bug in the library would affect both computations in the same way.

2. Reproducibility. We require replications to be reproducible, such that reviewers and readers can re-run the code and explore the work in detail. From that point of view, external dependencies are OK as long as it is straightforward to install and use the same version that was used in the replication. But the fewer dependencies, the better.

3. Understandability. A ReScience replication is not code, but a paper plus code that together explain a method to a human reader. Any dependencies must be understood by the reader as well, and therefore they should be either "scientifically simple" (meaning they embed little domain-specific knowledge, as for NumPy), or "well-known in their domain" (meaning that anyone competent enough to understand the paper at all would also be reasonably familiar with what the dependencies do).

4. Reusability. We'd like our replications to be good foundations to build further research on. This implies two desirable properties:
   a. Understandability again. You can safely reuse code only if you understand it well enough to be sure that it's correct for your own use case.
   b. Robustness of the code. If you want to build on some code, you want it to run on all relevant platforms, now and in the future.
   c. Generality of the code. The more general it is, the more variations are possible in the application.

Now let's compare the two approaches that have been evoked in this thread:

- "use NEST" wins on 4c and 4b is NEST is well maintained

- "write a special-purpose 200-line Python function" wins on 2 and 3/4a

As for 1, it depends on whether the original paper is based on NEST or not. We might not know for sure.

Next, what policy should ReScience adopt? My point of view is that we should be as strict as possible about independence, because that's where the real added value of replication work comes from. Among the other desirable properties, my personal priorities are 3, then 2, then 4, but that reflects my personal experience, so I'd expect others to disagree.

Finally, a comment on the argument put forward by @iampritishpatil and @apdavison that using external, well-tested, continuously maintained package is preferable. This is a common argument for software in general. And I support that argument for lots of software, be it Python, NumPy, Emacs, Firefox, etc. More precisely, I support that argument for software whose specification is much simpler than its implementation.

Take NumPy as an example. Anyone with some programming experience could re-implement a subset of NumPy for a given project. NumPy implements only well-known algorithms and data structures. Implementing them efficiently, consistently, and reliably in the Python context is a different matter. It's a difficult task, and best done by a handful of experts. But the NumPy user doesn't need to know about those technical details.

Now take NEST, or many other complex scientific packages. They propose sophisticated abstractions to their users that hide not only technical details, but also choices that are scientifically relevant. As a rule of thumb, if you are unable to implement a functionality of such a package yourself, you shouldn't use the canned implementation either. Of course people do, and this is a major cause of mistakes, including some that got much publicity recently (Geoffrey Chang's [retracted protein structures](http://doi.org/10.1126/science.314.5807.1856), for example, or the [fMRI false positive rates](http://doi.org/10.1073/pnas.1602413113)). And even if you are able to implement the functionality yourself, you probably ought to do it and compare with the canned version, just to make sure it really does what you think it does. I have seen a PhD student develop a depression after finding out that the two-years-worth of simulation data he had produced was worthless because of an undocumented approximation made in the code of a widely used molecular simulation package, because the approximation was inappropriate for his specific use case.

otizonaizit:
Thank you all for your comments. 

It is important to understand the discussion about the __current__ submission, that the code attached to it comprises mostly of plotting code, and lines referring to the NEST simulator in form of `nest.SetStatus(...)`, `nest.Simulate`, `nest.Connect`. Then the collected results are plotted. The functionality needed in NEST to simulate the models in this submission has been added to NEST somewhat recently, if @heplesser declines his review because he helped reviewing that code. So I am not even sure how much testing in the field this new functionality has gotten.

My issue is that I am not sure it is correct for ReScience to defer the reviewing of the code to the NEST community, and accept a submission that barely "uses" that code. As @khinsen clearly states above, using `numpy` is a different story, and we surely would not accept a submission that replicates some fancy integration algorithm by using on a single line a  just added new function to `numpy` without wanting to review the code in numpy too, wouldn't we?

While I think that it is acceptable to expect reviewers to install the dependencies needed to run the code in a submission, I don't think it is acceptable to ask them to review the code in those external dependencies. Or maybe it is? That is my question. And in this submission, really, there's otherwise no other code to review, because plotting and setting up is not where the dangerous bugs typically lie.

otizonaizit:
@gdetor: thank you!

khinsen:
I have opened [an issue on ReScience](https://github.com/ReScience/ReScience/issues/66) for the more general discussion of how to deal with shallow code delegating the grunt work to libraries. Please add your views there, leaving this space for the ongoing review!

otizonaizit:
@MSenden : in the introduction you state that you provide an implementation of the SG model by Gancarz & Grossberg and that this model has been implemented within the NEST simulator. It seems to me that the code needed to implement the SG model was very recently merged into NEST from [PR #858](https://github.com/nest/nest-simulator/pull/858).

As far as I can understand, this code is at the moment not available in any official release of NEST. 

After carefully evaluating the discussion here and talking with the editors-in-chief, we came to the conlusion that we would like to proceed with the review. In order to do that, I'd appreciate for you to either:

1. attach the needed code to this submission, and use it in NEST as an externally implemented module (I think it is possible in NEST to load externally defined models, right?), so that the reviewers have a chance to evaluate the quality of the implementation, or
2. explain in a bit more detail in the introduction which recent features of NEST are used (you are already mentioning some in the article at page 3), explicitly specify which version of NEST is needed to run the code, and if and why you consider this features tested well enough that the reviewers can trust them. 

I would like to thank you in advance for the additional work required. Your article is in my opinion very well written and I'd like to get it published.

@apdavison, @gdetor : are you OK with this way to proceed?

To everyone else in the discussion: we'll open an issue to discuss the general problem of domain-specific external dependencies and how to handle this sort of submissions in the future, so stay tuned at https://github.com/ReScience/ReScience/issues/66  :)

apdavison:
@otizonaizit Yes, that's fine with me.

gdetor:
@otizonaizit It works for me, too.

MSenden:
@otizonaizit 
Thank you for letting the review continue. We have been following the discussion ensuing our submission with interest. 

Following your recommendations, we have added the full code of nest v2.16.0-beta to the submission (within the new folder named `nest`). In this way the reviewers have a chance to examine the implementation more thoroughly as the implementation of every neuron model can be found in `../nest/nest-simulator-master/models`. 
A new section in the article (appendix A1) is devoted to give a brief overview of how rate-based neurons are implemented in NEST and refers to relevant files which reviewers might be interested in.
We did not add the code as an externally implemented module, however, as this would either be identical to the internal implementation and hence be redundant or differ from the internal implementation and hence not be fully representative of our work.

We also extended the introduction to provide more thorough background as to why we opted to use NEST. With regard to the point of why we consider recently added features tested well enough for readers/reviewers to trust them, we consider the internal review process of the NEST initiative very reliable. More important to the present discussion, however, we would argue that it is precisely through (numerous) successful re-implementations of existing models that the community can get an independent indication as to whether a software tool is reliable.

otizonaizit:
@MSenden : thank you for the changes! They look good and will help a lot in the review process!
@apdavison, @gdetor: you can start reviewing the paper (and the code :)))

rougier:
@otizonaizit Gentle reminder.

otizonaizit:
@apdavison , @gdetor : do you have a timeline for your review? :)

gdetor:
@otizonaizit I'm sorry for the delay. I'll try to finish the review by next week.

otizonaizit:
Hi @apdavison , are you still on this? Do you have an approximate timeline for the review? Thanks!

apdavison:
@otizonaizit yes, sorry for not replying sooner. I have installed everything and run the simulations; everything seems fine, so I now need to review the code. Unfortunately I have a major deadline for a grant in one week, so I won't be able to finish the review until the first week in April.

gdetor:
The authors reliably reproduced the results of [A Neural Model of the Saccade Generator in the Reticular Formation]. The source code runs smoothly and with no problems and the text is well-written.
The provided instructions for installing NEST on Linux are correct and overall I did not face any issue with their code or the simulations execution.  

General comments
------------------------
1. How the authors estimated the values from the figures of the original paper in order to compare against their own results? 
Did they use any software (like plot digitizer, http://plotdigitizer.sourceforge.net/) to extract the data points from the figures?
I think the authors should provide some more information in the text on how they compared their results against the original ones. 

2. The authors in order to integrate the tonic neurons ODE used the Euler-Maruyama method which is mainly used to integrate SDEs. Does this imply that they use some sort of noise in their implementation?   

3. Regarding the discrepancy:  What does it mean that it might stem from the input description (pg. 10)? In addition is not clear if F is indeed a function of time or some sort of constant. 

Text
-----
1. The text is well-written and makes clear all the details of the implementation. There are, however, some typos (see below). 
2. Please make clear that the subscripts l, r, u, d mean left, right, up and down. 

Typos: 
1. Introduction, pg 1, first sentence -> Commission
2. Introduction, pg 1, first sentence -> pan-European
3. Introduction, pg 1, line 6 -> collaboration
4. Methods, pg 2, line 7 -> Colliculus
5. Methods, pg 2, equation 2, there is extra parenthesis at the RHS third term. 
6. Results, pg 7, line 4 -> amplitude
7. Figure 7, pg 8 -> velocity
The authors might need to proofread the manuscript once again. 

3. In the Methods section, please add the proper punctuation in the equations. That would be nice if the authors write the ODEs in a more elegant form. 

Figures
---------
1. The x,y-labels as well as the ticks labels fonts are too small. Please make it larger. 
2. In addition, rotate y-axis ticks (abbreviations like Input, LLBN, etc) to match the original figures (easier to read).
Authors can use something like ax.set_ylabel('Input', rotation='horizontal', labelpad=20)
3. In the caption of Figure 4, did the authors mean staircase continuum instead of continued? 
4. In Figure 5, what is 0.02 and 0.1? Please provide more details regarding this figure. 
5. Figure 6, comments regarding the B panel are missing. 
6. In Figure 7, what do the dashed and the solid lines indicate?

Code
-------
1. The code provided in the setup_model.py script is well-commented.
2. The code in all the Exp?.py scripts need some more comments. For instance, t_relax variable, F, W, J.
3. Time vector T may confuse the reader (it's not immediately clear what is the relaxation time and what is the stimulation (simulation) one). 
4. The authors should provide platform information in README file.
See http://rescience.github.io/platform/
5. None of the scripts comply with the PEP8. Authors can use the autoflake tool to make their files comply with PEP8.
6. The code runs with Python 3 as well. The auhors need to add maybe the following lines in their 
Exp?.py scripts: 
        import sys
	if sys.version_info[0] < 3:
	  execfile('setup_model.py')
	else:
	  exec(open('setup_model.py').read())    

7. In Exp?.py scripts at line 45: (typos) is does -> does

apdavison:
I am satisfied that this submission is a full replication of the Gancarz and Grossberg technical report.

Installing NEST was straightforward. I adapted the Python scripts very slightly so they would run with Python 3 (replaced `execfile('setup_model.py')` with `from setup_model import *`). All scripts ran quickly and without problems, and produced visually identical figures to those in the article (there were some minor differences in the .eps files, presumably due to using a different version of matplotlib).

I skimmed through the C++ code that was added to NEST for this model, and studied two of the model implementations more closely, without finding any problems.

I think the code should be adapted to run with both Python 2.7 and 3.6 (Python 2.7 will reach end-of-life in about 20 months from now). As both I and @gdetor found, this is trivial to do using either `import` or `exec`. In general, I dislike the use of `execfile/exec` for code readability reasons: it makes it harder for the reader to understand where variables are defined, or to find their definitions, and it prevents IDEs or code editors from helping the user with this. Of course, `from X import *` is little better; the best approach would be to explicitly import into each script the required variables and functions from `setup_model.py`.

otizonaizit:
@apdavison @gdetor : thank you for your reviews!
@MSenden : could you address the reviewers' comments?

MSenden:
We would like to thank @apdavison @gdetor for taking the time to review our manuscript and their helpful comments. In what follows we provide a point by point response to the issues raised by the reviewers.

# Reviewer 1

## General comments

> How the authors estimated the values from the figures of the original paper in order to compare against their own results? Did they use any software (like plot digitizer, http://plotdigitizer.sourceforge.net/) to extract the data points from the figures? I think the authors should provide some more information in the text on how they compared their results against the original ones.

We are sorry for this shortcoming in the description of our methods. We are familiar with the plot digitizer and the original version of the manuscript indeed uses the tool to extract individual data points from figures 6, 9 and 10 of the original publication. The revised version of the manuscript now mentions this explicitly and describes in more detail the exact information we obtain from the three figures.

> The authors in order to integrate the tonic neurons ODE used the Euler-Maruyama method which is mainly used to integrate SDEs. Does this imply that they use some sort of noise in their implementation?

The reviewer is correct to point out that the Euler-Maruyama (EM) method is used to integrate stochastic differential equations. NEST is generally capable of handling noise (using the EE method by default and the EM method in the absence of passive decay). However, there is neither noise in the original model nor in our implementation of the saccade generator and hence the EM method reduces to the forward Euler method. The revised version of the manuscript now states that tonic neurons are integrated using the forward Euler method (rather than the EM method).

>  Regarding the discrepancy: What does it mean that it might stem from the input description (pg. 10)?

The most important implication of the discrepancy originating from the input description is that there are no issues with the model itself. Once the correct input is provided (the one matching the shape displayed in figure 10 of the original publication), the model produces the expected output. Furthermore, the fact that the discrepancy remains when the input is treated analytically implies that the deviation is not due to our implementation. In the revised version of the manuscript we have improved the corresponding section of the text accordingly.

> It is not clear if F is indeed a function of time or some sort of constant

F is a constant but we agree that the equations in the original version of the manuscript suggest that F is a function of time. In the revised version of the manuscript we have changed F(t) to simply read F.

## Text

> The text is well-written and makes clear all the details of the implementation. There are, however, some typos (see below). Please make clear that the subscripts l, r, u, d mean left, right, up and down.

Thank you for pointing out this omission. The revised manuscript now explains the subscripts. Furthermore, all typos found by the reviewer have been corrected and the manuscript has been thoroughly re-read to check for further typos.

> In the Methods section, please add the proper punctuation in the equations.

For the revised version of the manuscript we have rechecked all equations.

> That would be nice if the authors write the ODEs in a more elegant form.

The original version of the manuscript uses \frac{d}{dt} to indicate temporal derivatives in line with the style of the work we refer to, the revised manuscript uses \dot{\boxempty} to reduce visual clutter.

## Figures
> The x,y-labels as well as the ticks labels fonts are too small. Please make it larger

We have adjusted the figures accordingly.

> Rotate y-axis ticks (abbreviations like Input, LLBN, etc) to match the original figures (easier to read).

Thank you for the advice, we have done that.

> In Figure 5, what is 0.02 and 0.1? Please provide more details regarding this figure.

Sorry. These values are arbitrary and have been removed from the figure

> Figure 6, comments regarding the B panel are missing.

This is a misunderstanding, the respective sentence in the original version of the manuscript is not specifying the panels but just stating an observation. In the revised version we have improved the description of the panels and moved the respective sentence to the main text.

> In Figure 7, what do the dashed and the solid lines indicate?

The reviewer is right that this is not obvious in the original version of the manuscript. The line styles  differentiate between low and high velocity saccades. The revised manuscript makes this explicit.

## Code
> The code in all the Exp?.py scripts need some more comments. For instance, t_relax variable, F, W, J.

We have added further comments.

>Time vector T may confuse the reader (it's not immediately clear what is the relaxation time and what is the stimulation (simulation) one).

The newly added comments explain this better.

> The authors should provide platform information in README file. See http://rescience.github.io/platform/

Thank you for pointing this out. We have added this information.

> None of the scripts comply with the PEP8. Authors can use the autoflake tool to make their files comply with PEP8.

The scripts now comply with PEP8.

> The code runs with Python 3 as well. The auhors need to add maybe the following lines in their Exp?.py scripts: import sys if sys.version_info[0] < 3: execfile('setup_model.py') else: exec(open('setup_model.py').read())

Both reviewers are correct to point out that the code runs with Python 3. This is now explicitly mentioned in the manuscript. In compliance with the second reviewer, we have opted for using `from setup_model import *` to set up the model in each script.

> In Exp?.py scripts at line 45: (typos) is does -> does

We corrected this typo.

# Reviewer 2

> I think the code should be adapted to run with both Python 2.7 and 3.6 (Python 2.7 will reach end-of-life in about 20 months from now). As both I and @gdetor found, this is trivial to do using either import or exec. In general, I dislike the use of execfile/exec for code readability reasons: it makes it harder for the reader to understand where variables are defined, or to find their definitions, and it prevents IDEs or code editors from helping the user with this. Of course, from X import * is little better; the best approach would be to explicitly import into each script the required variables and functions from setup_model.py.

The code indeed runs with Python 3. This is now explicitly mentioned in the manuscript. Thank you for pointing out the option to use `from setup_model import *` which we now use in our scripts.

rougier:
@otizonaizit @gdetor @apdavison Gentle reminder...

otizonaizit:
@gdetor @apdavison : my impression is that the authors have addressed all your points in great detail, so I think I just need a confirmation from you that I can accept the submission as is :)

apdavison:
@otizonaizit I was planning to re-run the scripts just to double check there have been no regressions, but otherwise I'm happy for you to accept the submission.

gdetor:
@otizonaizit  The authors addressed properly all the issues. All the scripts run now smoothly and out-of-the-box. 
However, there are two more typos in the text: 
1] Methods section, pg. 2, first paragraph, sentence: "Please note that ..., respectively". The subscripts are not in the proper order.
2] Methods section, pg. 3, last paragraph:  forward Euler method -> Forward Euler method

I suggest to you to accept the submission.

MSenden:
@gdetor thanks for endorsing the article. We corrected the typos you mention.

rougier:
@otizonaizit Any update?

otizonaizit:
Article is published with [DOI](https://doi.org/10.5281/zenodo.1241004)

PR #48, opened 2018-03-13
Add function 2 and 4

QileWu:


rougier:
@QileWu Is it a "too early" submission?

rougier:
I guess this mean yes 😄

Issue #16, opened 2016-04-18
Re-implementation of open source methods in another language


arokem:
Dear Rescience editors. In the course of our work, we have created a Python implementation of a method that was previously available as open-source R code. Is this implementation within the scope of Rescience? Thanks! cc:@kpolimis, @bhazelton

rougier:
Do you have a reference paper to target for the replication ?

arokem:
Yes. This is the paper: http://jmlr.csail.mit.edu/papers/volume15/wager14a/wager14a.pdf, and the previous implementation: https://github.com/swager/randomForestCI

rougier:
It is okay as long as you do not make a simple "translation" of the R code. The idea of the replication is really to check if the original article is self-sufficient when describing method or model (i.e. without the accompanying code) or if some information is incorrect or missing. In the end, the original article + your article should be sufficient for future replications.

@khinsen What do you think ?

FedericoV:
This seems like a better fit for a project like
http://contrib.scikit-learn.org/ to me.

rougier:
Didn't know this. But anyway, you can do both actually (publication and contribution).

khinsen:
@rougier I agree: the main point of ReScience is doing replication in the sense of writing a new implementation that should produce results identical to published ones. If a published implementation already, we should ask for a "clean-room reimplementation" although this can of course not be verified.

In my personal experience, a second independent implementation is a great way to find mistakes (in _both_ implementations), so I am tempted to suggest that we even encourage that kind of submission for ReScience.

FedericoV:
There's an additional issue as well: most code in R is licensed GPL, while
most Python code is licensed MIT.  If the code is a clean room
implementation, you can use MIT/BSD as a license, while if it is a
derivative work of the R code, you have to use it as GPL, which limits its
use within Python.

rougier:
Replication is not a derivative work for me.

FedericoV:
I am not a lawyer, but I believe that if you look at GPL code while you
implement the Python code, it counts as derivative.

arokem:
Might be generally of relevance, but not for this particular case. We did
not do a "clean room" implementation (IIUC what that would entail).
Instead, we looked at the original code, but in this case, it is under an
MIT license.

rougier:
[...]

PR #57, opened 2019-04-10
Review Request : R. Larisch 

rLarisch:
**AUTHOR**

Dear @ReScience/editors,

### Original article

**Title:** Connectivity reflects coding: a model of voltage-based STDP with homeostasis 
**Author(s):**  Clopath, C., Büsing, L., Vasilaki, E. and Gerstner, W.
**Journal (or Conference):** Nature Neuroscience 
**Year:**  2010
**DOI:**  10.1038/nn.2479
**PDF:** https://www.nature.com/articles/nn.2479  

### Replication

**Author(s)**:   Larisch, R.
**Repository**:  https://github.com/rLarisch/ReScience-submission/tree/Larisch-2019
**PDF**:  https://github.com/rLarisch/ReScience-submission/blob/Larisch-2019/article/Larisch-2019.pdf
**Keywords**:  Spike Timing-Dependent Plasticity (STDP), Computational Neuroscience, Unsupervised Learning
**Language**: Python

### Results

### Potential reviewers

---

**EDITOR**

* [x] Editor acknowledgment (@gdetor)
* [x] Reviewer 1 (@cJarvers, 27 May 2019)
* [x] Reviewer 2 (@apdavison, 27 May 2019)
* [x] Review 1 decision [**accept**]
* [X] Review 2 decision [**accept**]
* [X] Editor decision [**accept**]


rougier:
Thanks for your submission and orry, I'm very late. I'll assign an editor soon.

rougier:
@otizonaizit Can you handle this submission?

rougier:
@rLarisch Very sorry for the delay but we've been in the process of redesigning the website and the publication process. Things should be back to normal. Note that if you want (not mandatory) you can convert your submission to the new template design.

@gdetor Can you edit this submission?

rLarisch:
Hello,

@rougier thank you for your response. 
Changing to the new design should be possible. Only one question: Should I just download the new LaTeX template and recreate the PDF file in the article directory, while the rest stays at it is?

rougier:
Yes, but it is really up to you. The new template is latex only (no more markdown because we had too many problems). There is also an overleaf available if your prefer. The new submission process does not require you to first clone (and then PR) but since you've already submitted, we can leave it as it is (I mean for the submission).

gdetor:
@rougier Sure

cJarvers:
I'd be happy to review this.

gdetor:
@cJarvers  Ok.
@apdavison Could you review this submission?

apdavison:
@gdetor yes, happy to review this

rLarisch:
Hello everyone,

sorry for the delay,
in the 'article' directory you will now find a new sub-directory called 'new_template'. It contains the article.pdf created with the new ReScience Latex template and the metadata.yaml file. 
During the review process, I will update the old and new article.pdf-files.

Looking happily forward to the progress in the review process.

rougier:
@gdetor Gentle reminder

apdavison:
@rLarisch I tried to install ANNarchy on MacOS as follows:
```
conda create -n review-rLarisch python=3 numpy matplotlib cython sympy scipy pip
conda activate review-rLarisch
pip install ANNarchy
```
but I get the following error:
```
Building wheel for ANNarchy (setup.py) ... error
  ERROR: Complete output from command /Users/andrew/miniconda3/envs/review-rLarisch/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/private/var/folders/2k/mhzyfkfs7h76v3pfyjbksb540000gq/T/pip-install-iqh87f61/ANNarchy/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/var/folders/2k/mhzyfkfs7h76v3pfyjbksb540000gq/T/pip-wheel-9xvvzipp --python-tag cp37:
  ERROR: Checking for setuptools... OK
  Checking for numpy... OK
  Checking for cython... OK
  Checking for CUDA... NO
  Warning: CUDA is not available on your system. Only OpenMP can be used to perform the simulations.
  Installing ANNarchy 4.6.8
  	Python 3.7 ( /Users/andrew/miniconda3/envs/review-rLarisch/bin/python )
  	Includes: `/Users/andrew/miniconda3/envs/review-rLarisch/bin/python3-config --includes`
  	Libraries: -L/Users/andrew/miniconda3/envs/review-rLarisch/lib
  	Cython: cython
  	Numpy: /Users/andrew/miniconda3/envs/review-rLarisch/lib/python3.7/site-packages/numpy/core/include
  ANNarchy/core/cython_ext/Connector.cpp:637:10: fatal error: 'ios' file not found
  #include "ios"
           ^~~~~
  1 error generated.
  error: command 'gcc' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for ANNarchy
```

rLarisch:
Hello,

@apdavison I contacted the developers of ANNarchy and they need some information about your system: 
What is your MacOS version?
Which c++ compiler (and version number) do you use?
What is the version number of the Cython package in your conda environment?

apdavison:
thanks for the quick reply @rLarisch 

macOS 10.14.2
```
$ g++ --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

```
$ conda list cython
# packages in environment at /Users/andrew/miniconda3/envs/review-rLarisch:
#
# Name                    Version                   Build  Channel
cython                    0.29.10          py37h0a44026_0
```

khinsen:
Note that `g++` (and also `gcc`) are thin wrappers around `clang` under macOS. That's what it says in the description if you read it carefully, but many people miss it and believe they are using GCC.

vitay:
Sorry to intervene here, but it may be faster.

clang on MacOS 10.14 dropped support for the STL library libstdc++, to be replaced by libc++ :

https://github.com/cython/cython/issues/2694

I have uploaded on pip a new version 4.6.8.1 that should now work on newer OSX versions. Please tell me if it fixes your problem.

Perhaps @rLarisch should update the minimal required version for ANNarchy in the README.

apdavison:
@vitay No need to apologise, many thanks for the help. I think the fact you could easily do that is one of the big advantages of using Github for reviewing.

I installed the new version, and it fixes the problem.

gdetor:
Hi, @cJarvers @apdavison Any progress with the review?

cJarvers:
I ran the code without problems and read the paper. I still need some time to look through the code in detail. Since the explanations in the paper are quite detailed, I hope this won't take too long.

apdavison:
I also ran the code, on Linux and MacOS, but there are some differences in the figures that are generated. I will post a more detailed report next week.

gdetor:
Hi @cJarvers @apdavison, gentle reminder.

apdavison:
## Review of replication by René Larisch of "Connectivity reflects coding: a model of voltage-based STDP with homeostasis"

### Re-running the simulations

I re-ran the simulations with the provided code, but with somewhat more recent versions of the libraries than mentioned in the manuscript (Python 3.7.3, ANNarchy 4.6.8.1, matplotlib 3.1.0, numpy 1.16.4). Of the figures generated by the simulation, Figures 1, 2 and 3-Right looked the same as in the manuscript, but Figures 3-Left and 4 were qualitatively different - see attached figures. It is understandable that details of figures can change between implementations, especially where random number generators are used, but I would expect the code provided to produce identical results to those shown in the manuscript.

Fig3_rateCode

Fig4_stableW

Fig4_RF


### Replication of the original model

This reimplementation qualitatively reproduces the findings of the original study.

There are quantitative differences with respect to the original results, mostly small, and partly explained in the manuscript, although the reasons for some of the differences were not isolated, and there was no systematic study of how the order of numerical operations affected the results. In the section "Non-reproduced experiments", it is unclear whether the experiments could not be reproduced or whether reproduction was not attempted.

Given this, I would characterize this as a partial, not full replication. I nevertheless support publication because:

1. it is understandable that a reimplementation using an existing simulation tool such as ANNarchy does not allow an exact replication of the original results due to differences in the order of numerical operations, and this study is valuable as an illustration of the limitations in numerical replication and of the trade-offs in using more powerful, time-saving tools such as ANNarchy.

2. the essential principles elucidated in the original study are preserved in the replication despite the quantitative differences, which is evidence for their robustness.

For this to be characterized as a "reference implementation", I think a clearer and more quantitative dissection of the differences would be needed. As a minimum, the Figures should, where possible, plot the MATLAB and ANNarchy results on the same axes to show how large the differences are. For Figure 3, I do not see any justification for using a colour scheme that is "similar" to the original publication. The identical colour scheme should be used to avoid any risk of camouflaging differences.


### Minor issues

* In Tables 1 and 2, it would be helpful to the reader if both the original and changed values were given.
* In the tables and at a few points in the text, physical quantities are incorrectly typeset - units should be in an upright (not italic) font, and there should be a thin space between number and units (https://tug.org/TUGboat/tb18-1/tb54becc.pdf is a very helpful guide).
* Similarly, in equations subscripts like "ref" should be set in upright type.
* At several points, "Python" is written "python".
* In the section "Parameter setup", 2nd paragraph, the sentence "This forced us to change the value of some parameters or some of the simulations" is unclear. Perhaps "or" is supposed to be "for"?
* In the section "Experiment descriptions", 7th paragraph, there is a very unclear sentence which should be revised: "the recurrent connectted neurons receives input from neurons, they spiking 20 ms one after another to realize a temporal spiking order of the ten recurrent neurons".

cJarvers:
# Review of replication by R. Larisch, [Re:] Connectivity reflects coding: a model of voltage-based STDP with homeostasis

This review is based on the branch Larisch-2019, commit 2acfe30e661090a25fada6f26416c0e87e8dba4f.

## Replication

Several key results from the original paper are reproduced. As @apdavison already stated, there are some minor quantitative differences, which were due to implementational differences explained in the paper. While I think the explanations could be more detailed (see below), the results are sufficiently similar to count as a replication in my opinon.

Several of the results in the paper were not replicated. I concur with @apdavison that the author needs to make clear whether the replication failed (in which case the results should be included) or  was not attempted and that this should count as a partial replication.

Generally, I think the label "partial replication" in its current form is a little confusing (as discussed in ReScience/ReScience#67). Maybe a distinction between partial scope (i.e., some experiments not attempted) and partial success (i.e., tried to replicate all experiments but failed with some) may be useful.

The experiments that were not replicated are (1) the simulation of a network with 10 excitatory and 3 inhibitory units (Fig. 5 in original paper), (2) the experiment with moving input patterns (Fig. 6 in original paper), (3) the receptive field training with different rates (Fig. 7e in original paper) and (4) the control experiments for learning with a classical STDP rule (Fig. 4c and 4d in original paper). The reasoning for omitting the first two is explained in the paper (essentially, that given the reproduction of the other experiments, these two are redundant). I understand the author's reasoning and agree that these experiments could most likely be replicated given the success with replicating the other experiments. However, one should not trust intuition in a case like this; there may still be some problem in the original experiments that renders them non-replicable.

Concerning the other two non-replicated experiments, I suppose they are ignored because they are essentially control conditions and not part of the main narrative of the paper. Nevertheless, a full replication should include them. For example, a central claim of the original paper was that the patterns in Fig. 4 that could be produced with their model cannot be reproduced with a standard STDP learning rule, which they show with Fig. 4c and 4d. If this was due to some error in their implementation, this would undermine a key conclusion.

Including these additional replications would be considerable extra work and the implementation in its current form is already valuable. Therefore, I support publication as a partial replication, even though I would prefer to see the remaining experiments replicated as well.

One other potential problem is that the original Matlab code was used as a basis for the reimplementation. If I understand the discussion in #16 correctly, this is discouraged since it increases the risk to port errors from the original into the replication without realizing. I think the code in this reimplementation is independent enough to warrant publication, especially since the paper makes it quite clear which parts were taken from the Matlab code (however, I was slightly confused by the statement that equation 1 was derived from the Matlab source, since it was given in the appendix of the paper). @gdetor, can you or one of the other editors clarify what the threshold for dangerous overlap between original and reimplementation is?

Finally, the receptive fields produced when I run Fig4_RF.py are much less regular than the ones in the original paper or in this replication (sample below). I have no idea what the cause of this may be. Any ideas, @rLarisch?

Fig4_RF

## Reproducibility

I ran the code with Python 2.7.16 (numpy 1.16.4, matplotlib 2.2.4, scipy 1.2.2, sympy 1.4, Cython 0.29.13) and Python 3.7.4 (numpy 1.17.0, matplotlib 3.1.1, scipy 1.3.1, sympy 1.4, Cython 0.29.13).
I tried ANNarchy versions 4.6.6 and 4.6.8.1 on both Pythons.
The results were identical in all settings, except for minor differences in how the plots were rendered (axes etc.) and the fact that the experiments that involve randomness did not have fixed seeds (see below).

One note regarding versions: the version information given in the paper (Python 3.4 or later, ANNarchy 4.6.6, numpy 1.11.0, matplotlib 1.5.1) differ from the versions specified in the repository's README file (Python 2.7, ANNarchy 4.6.4, numpy and matplotlib same). It would be good to have these identical. Generally, presenting the version information 4 times (in two places in the paper, in the root README and in the code README) with different levels of granularity can be confusing.

The results are not exactly reproducible, since each run uses different random seeds. As @apdavison notes, this means that figures 3 left, 4 left and 4 right cannot be reproduced exactly, as these depend on random sampling. In order to get exactly reproducible results, I recommend to set a fixed seed for each simulation. According to the ANNarchy documentation (section 2.12.2) this can be
done with `setup(seed=VALUE)`.

## Clarity of the code

The code is well structured, clear and well commented. Some of the scripts for experiments still contain some code remnants in comments, which should be removed. These are: 

- Fig1_clamp.py in lines 26 and 156-158
- Fig1_window.py in lines 94, 101, 117 and 119

In addition, Fig2_burst.py still contains debug print statements in lines 134 and 179, which could be removed.

There are two implementations of the neuron dynamics, net_fix.py and net_homeostatic.py. The only difference seems to be how large the additive constant is that offsets the membrane voltage 2ms after a spiked during the resolution trick. In one case it is 51.75, in the other it is 49.5. I don't see where this difference is coming from. @rLarisch, can you clarify? Given that the files have such a large overlap, it may be better to use just one file and interpolate the value into the string given an input argument or something similar.

My main problem with reading the neuron implementation was that a lot of the parameters seemed to come out of nowhere. It might be helpful to have a table with all parameter values somewhere in the paper. It could also include a brief statement about what the meaning of each parameter is.

Some more specific issues:

- In Fig1_clamp.py, the input spike train is supposed to have 25Hz over 50ms, which should correspond to 1250 spikes. However, it only generates 200 spikes, which correponds to 4 Hz. Similarly, line 121 seems to generate input with 10Hz instead of 100Hz.
- In Fig1_clamp.py, I am slightly concerned about line 133. Here, the results of the cortical experiment are essentially clipped to the maximum value of the hippocampal experiment. If this step is left out, they grow much larger. The results in the paper are also upper-bounded, but I could not find out whether they also clipped the values like this. If that is the case, it should definitely be stated in the paper, as it changes the interpretation of that figure.
- In the original paper, the pairings for the burst experiment were repeated 60 times. If I read Fig2_burst.py correctly, each pairing is only executed once. Is that correct? If so, why are the relative weight increases the same as after 60 pairings in the original paper? If not, where is the 60-fold repetition that I am missing?

## Clarity of the paper

I think the paper may benefit from some restructuring. Currently, the description of the model is followed by the section "Experiment descriptions", in which each experiment is mentioned, followed by detailed explanations on the implementation, followed by the experimental results. This leads to a kind of back and forth that makes it hard to keep all details in mind. I had to scroll back and forth a lot. I think it would be better to have the explanation of the model directly followed by the implementation section. This could be followed by a description of all experiment, with the details about how the code was adapted directly followed by the results.

In addition, I found some minor problems that made it harder for me to understand the paper right away:

- It may be better to put the equations for the synaptic model in one block. As it is, equation 6 is surrounded by text that does not belong to it, which briefly tripped me up while reading.
- The reset conditions of the model are not shown in equations. Some explanations are provided in the text, but having one concise summary of which variable is reset to what value after a spike would be useful.
- In the explanations following equation 7, it is mentioned that `u_-` has to be above `theta_-` for LTP to occur. I think this should be `u_+`, since `u_-` has not been introduced at this point.
- In the explanations following equation 9, I am not sure where the `w_max` is coming from. Nor do I understand how the value for it was chosen (table 1).
- In the description of the voltage clamp experiment (p. 4), the phrase "is changed from a fixed value of -80mV to 0mV" is misleading in my opinion. Other formulations like "is set to a fixed value" may be better. The following sentence "Resulting changes in the learning rule are implemented as mentioned   in the original publication" is also confusing. Does this refer to the equation in the online supplement? If so, I am not sure what changes to the code would be necessary.
- In the description of the burst timing experiment, it is not quite clear that the 50Hz postsynaptic rate refers to the frequency during the burst and that the neuron is quiet otherwise.
- For the RF learning experiment, the maximum firing rate is set to 50Hz instead of the 37.5 Hz used in the original paper. This change is stated but not explained.
- I took a while to understand that Fig. 4 left refers to the new Poisson pattern experiment and is not supposed to be an equivalent of Fig. 7c in the original paper. This could be made clearer, maybe by also replicating Fig. 7c. In addition, the color map used in Fig. 4 left is not linear. Using either greyscale or a linear color map like viridis would be better in my opinion.

## Misc

- The brackets in equation 9 do not seem to have any purpose. In the code, a `pos` is applied to the difference term, which is not given in the equation.
- Tables 1 and 2 present the changed parameters. In the text they are referred to as "the tables below", but actually appear above that bit of text. Referring to them by numbers may be better.
- Only the changed parameters are shown in a table. It would be helpful to have a table with all the other parameters as well. While these are given in the original paper, not all readers may have access to Nature and it's not open access. That table could appear in an appendix.
- On page 7, `neuron_eqs` runs into the margin.
- The paper states that the RF experiment was run on 200000 patches, whereas the code actually runs 300000.

gdetor:
@apdavison @cJarvers Thank you for your reviews.
@rLarisch Can you address the reviewers' comments?

rLarisch:
@apdavison @cJarvers At first, thank you for your reviews and comments.
I will try to answer all of the major comments properly:

1. To the reproducibility of Fig. 3 - left and both figures in Fig.4:
1.1. I wasn't aware that the code should reproduce 100% the same figures as in the manuscript. As already mentioned by @apdavison , the weights are initialized randomly and because of that the results change and I thought it is import to create qualitative similar results with different initial conditions to prove that the implementation is stable. But I can rerun the code again with a fixe seed to generate always the same results and switch the images in the article with the new images.
1.2. However, this is not an excuse for the strange-looking receptive fields. As far as I looked in the python scripts, I accidentally mixed up different versions for the net_homeostatic.py. I'm really sorry about that! I actually looking for the right implementation to get proper receptive fields.

2. The non-reproduced experiments:
2.1 As @cJarvers mentioned, I did not try to replicate Fig.4.c and Fig.4 d of the original manuscript, because they are created with a "normal" STDP learning rule and so I ignore them to concentrate on the reimplementation of the Clopath et al. (2010) rule. But it should be not a problem to implement a "normal" STDP rule and redo the experiments.
2.2 Fig.5 and Fig.6, I'm aware of the problem that only by redoing the experiments it is possible to show the correctness of the reimplementation. As mentioned by @cJarvers , it would take a lot of work to implement and test the protocols. So if it is possible, I would still not reimplement both experiments.

3. Use the Matlab code as a reference:
In the original manuscript, at equation 3 in the Methods section, you can find the term A_{LTD}(\\\u) (sorry I do not know, how I can create a good looking equation here). Below that you can find the definition of this term. One of the secrets of the original manuscript is, how the \\\u is calculated. It is a part of the homeostatic mechanisms, what is necessary for many experiments (Fig.4, Fig.5, Fig.6, and Fig.7). The Matlab code given in the appendix of the paper did not contain the equation for the \\\u. So it would be only possible to reproduce Fig.1, Fig.2, and Fig.3. So I used the provided Matlab implementation from modelDB, where all necessary equations are given (see RFdevelop.m in the source files on modelDB). Besides that, I understood the risk of import coding errors, as even the provided code of modelDB is inconsistent with the manuscript. In the provided Matlab code, a fix spiking threshold is used instead of the adaptive threshold, as mentioned in the paper (where my reimplementation differs from the modelDB Matlab code because I use for all experiments the adaptive spiking threshold). 
I'm aware of these inconsistencies, but find it hard to find clarity in the provided Matlab codes. 
I would add an additional paragraph to the manuscript to write more about the problems with the Matlab code in the original appendix and about the inconsistencies in the provided Matlab code on modelDB.

4. Differences between Matlab and ANNarchy implementation:
For the "stable weight task" (see Fig4.-left) I can plot the result from the Matlab code. Nonetheless, both codes working with randomly initialized weights and randomly creates input patterns. I can create a modification of both implementations, where weights and input pattern are loaded from an external *.mat file to ensure that both implementations are initialized with the same weights and get the same input presented and report the results of the experiment of both implementations. I would add them into the appendix.

5. Different mentioned version numbers in manuscript and Readme.md files:
The information about the necessary packages should be consistent. I will fix this.

6. Differences in net_fix.py and net_homeostatic.py:
The mentioned difference in the offset of the membrane potential should not be there. It should be in both version +49.5. Despite that, the main difference between the files is found in 'ltdTerm = ..' in the 'equatSTDP' string to define the equation for the STDP learning rule. In net_fix.py you can find the "vmean/urefsquare" and "vmean" is only a fix value, what is necessary for some of the experiments (as mentioned in the submitted article, page 5, first paragraph and fourth paragraph ). In net_homeostatic.py you can find "post.vmean/urefsquare". The "post.vmean" address to a variable called "vmean" defined in the neuron model and depend on the membrane potential. This implements a homeostatic mechanism, which depends on the postsynaptic membrane potential. The separation in to two files should make it easier for the reader to know, which version is used in which experiment is used. If it confuse the reader, I will put the two network definition in one file. 

7. Origin of the parameters:
Most parameters are taken from the original publication (as mentioned in the submitted article (below equation 4 and below equation 10)). Changed parameters are given in the tables and found experimentally. If it is ok, I would add a table with the parameters reported in the original publication in the appendix right after the literature.

8. Comments to the voltage-clamp experiment :
8.1. The 25 Hz over 50 ms is a mistake in the submitted article manuscript. Right are 25 Hz, respectively 100Hz over 50 seconds, not milliseconds.
8.2. In the original publication, they do not mention if they clipped the weights there or not. The learning rule is not able to archive a stable value without the help of a hard upper bound. The blue curve in Fig.4 h in the original publication (or Fig.1. right in the submitted manuscript) is only possible with a hardbound at the right value. Because of that, I assume that in the original publication they used a hard upper bound too, respectively clipped the blue line. 

9. Spiking Burst - Protocol: 
It is correct that the protocols are done only once. The two reasons why the relative weight values are the same: First, the time between two pairing repetition is 0.1 Hz ( or 10 seconds). In this time, all variables in the pre- and postsynaptic neurons have time to get back to their initial values (please mention, that the rese() statement only set the neuron variables). Second, the pre- and postsynaptic neurons are stimulated externally. As a consequence of the learning rule, the weight change is constant for each of the 60 paring repetitions and depends only on the time between a pre- and a postsynaptic spike, not on the synaptic weight between them. The relative relation to the initial weight depends only on the value chosen for the initial weight. 

10. Minor points:
I would fix the grammar and linguistic mistakes as mentioned by both reviewers and put the implementation of the model directly under the description of it to improve the readability of the article.

I hope that I adressed all major questions and comments.

It is not that easy to reimplement this model, so I would be fine if this publication is a partial replication. 
@gdetor I have not that much experience with the review process in ReScience, is it ok if I change it from full to partial replication in the initial comment by my self ?

@gdetor 
Because I have a deadline for another paper in two weeks, I want to ask for some time (around three weeks) to implement the changes in the article and in the code ?

Again, thanks @cJarvers and @apdavison for the comments.

gdetor:
@rLarisch Sure, no problem about taking more time to address reviewer's comments. Regarding the partial replication, I would suggest making first all the necessary changes and addressing the rest of the reviewer's comments and later I'll change the tag to partial replication.

gdetor:
Hi @rLarisch, gentle reminder.

rLarisch:
Hello @gdetor,  @cJarvers, and @apdavison I have updated the article manuscript and source codes. 
Now the scripts should always produce the same results (thanks to seeds). 
I rearranged the text so that it is hopefully better to read.
Now, the code directory contains only one 'network.py' and an additional variable to switch between a fix and a dynamic homeostatic mechanism ('set_fix'). 
Some grammar, spell and content mistakes are fixed.
Add a supplementary.pdf where you can find for two tasks with random weight initialization some of the results to look how robust is the reimplementation.
Add some more explanation on the text to hopefully clear things up.
The old markdown as the new Latex Template is updated.
For the experiment to investigate the influence of input activity and temporal spiking order, I repeated this experiment with the Song and Abbott (2001) rule. You can find the results on Fig.3.
And reproduce now the experiment, which analyzes the influence of the input firing rate on the size of the Gabor like receptive field (Fig.4.c).
For the normal receptive fields (Fig4 b), only the field in the upper left corner will be reproduced if you let the Python script run. The other three are with randomly chosen weights. 

I'm open for questions, comments and suggestions.

Best regards,
René

gdetor:
@rLarisch Thank you for addressing reviewers' comments. 
@cJarvers @apdavison Can you please review the new submission?

gdetor:
@cJarvers @apdavison Gentle reminder

rougier:
Any update ?

cJarvers:
Sorry for the delay. Here's my review of the updates, based on commit c869eee56b232f691fcea93a89b48e87d59d5b47:

All of my main concerns have been addressed by the updates. Concerning **replication**, the control experiments (Figure 4 c and d) are now included. It is now much clearer which parts of the implementation were taken from the published Matlab code and how and why certain parameters were changed. Figure 5 and 6 are still not reproduced, but as discussed before I do support publication as a partial replication.

The code is now **reproducible** thanks to the use of a random seed in the experiment on receptive field formation.

Concerning **clarity**, I think the paper reads more fluently now. The supplement helps keep an overview over the parameters.

Some minor issues remain, which should be easy to fix:

- Section 2 of the supplement is empty in the pdf, since the figures are placed on another page. This makes for a weird visual impression. One or two sentences referring to the figures would help.
- Figure 4a and supplementary figure 2 should better be in greyscale or at least in a linear colour map. The current use of `jet` makes the figure harder to interpret in general and makes it difficult to read for colourblind people (for a visualization of the problem, see for example [this blogpost](https://agilescientific.com/blog/2017/12/14/no-more-rainbows).
- The plot of the weights in Figure 4a (Line 138 in `Fig4_RF.py`) seems to mistakenly use `plot` instead of `imshow`, which leads to a strange result for `weightsRF.png`. 
- Several of the equation references did not get resolved in compilation in the pdf generated with the new template.
- The references to the Nevian and Sakmann (2006) paper on page 10 still have a `@` prefixed to them.
- The section on reimplementation (page 5) still mentions the two distinct network files `net_fix.py` and `net_homeostatic.py` even though the two have been combined.
- In the pdf generated with the new template, the code blocks have a strange alignment. I guess this is an issue with the template itself.
- In the section on the voltage clamp experiment (page 8) `theta` is spelled out and not written as the Greek letter.

rLarisch:
Hello @gdetor and @cJarvers,
I have updated the article manuscript to fix the minor issues mentioned by @cJarvers and add the Python protocols with the standard STDP learning rule to the "startAnalysis.py" file.

Still looking forward to the comments of @apdavison.

Best regards,
René Larisch

gdetor:
@cJarvers Thank you for the review. 
@rLarisch Thank you for addressing the comments. 
@apdavison Gently reminder.

apdavison:
I am happy with the changes, in particular running the code locally now produces identical figures to those in the manuscript.

A couple of minor points:
- as @cJarvers pointed out above, In the section on the voltage clamp experiment (page 8) `theta` should be `\theta`
- the overview mentions Python 3.6, the "Reimplementation" section Python 3.4.

As I mentioned in the original review, I don't think this should be described as a "reference implementation" because of the parameter changes that were needed, but I support publishing it as a partial replication.

gdetor:
@apdavison Thank you for the review
@rLarisch Could you please address reviewers' comments?

rLarisch:
@apdavison, @cJarvers

I have fixed the theta in both PDF-files and set the Python version consistent in both files.

@gdetor, it would be fine for me to publish it as a "partial replication", If it is ok, I would change it in the original Request from "fully replicated" to "partially replicated".

rLarisch:
Hello @gdetor,
I have seen that in the Header of the old ReScience template stands: "A reference implementation of ". Now that the submission is only a partial replication, to what should I change the header? And where?

rougier:
You can explain in the introduction that this is a partial replication of ...

gdetor:
@apdavison and @cJarvers Are you ok with the final draft?
Thank you

apdavison:
@gdetor yes

cJarvers:
@gdetor yes

rLarisch:
@gdetor @rougier I add a sentence at the end of the introduction about the fact that this is only a partial replication.

gdetor:
@rLarisch Congratulations, your paper has been accepted. I started the publication process, however, I noticed that the license is missing from your home repository. Can you please update the repository?
Thank you @apdavison and @cJarvers for your reviews and valuable comments.

rLarisch:
@gdetor A LICENCE.txt to the repository. Thank you for your help during the process.

@apdavison and @cJarvers thank you both for your reviews and your constructive comments.

gdetor:
Hi @rLarisch, because you use the old Rescience template you have to update the metadata on your side. I attach the corresponding metadata.yaml file that you have to use (I have filled out all the necessary information). 
[metadata.txt](https://github.com/ReScience/ReScience-submission/files/3821568/metadata.txt)
Don't forget to rename the metadata.txt to metadata.yaml
Thank you.

rLarisch:
Hi @gdetor, there was a metadata.yaml file in the article/new_template/ directory. I have now paste your metadata.yaml file in the article/ directory and have updated the .yaml file in the article/new_template/ directory. I hope this is ok.

gdetor:
Hi, @rLarisch thank you for updating the repo, however, you have to regenerate the pdf (new template) using the metadata file I sent you since it has all the necessary information for publishing the article. Thank you.

rLarisch:
Hi @gdetor, I have updated the article.pdf from the new template. Should I update the corresponding informations in the *.md file from the old template too ?

gdetor:
@rLarisch It would be nice to have it for consistency. Furthermore, I would like to ask you to regenerate the pdf with the following metadata file (I confused the sandbox metadata with the Zenodo one). The attached one is the final. Thank you and my apologies for the inconvenience. 
[metadata.txt](https://github.com/ReScience/ReScience-submission/files/3832973/metadata.txt)

rLarisch:
Hi @gdetor, the article.pdf with the new metadata.yaml file is updated and the corresponding information in the old *.md entered and the corresponding pdf fresh compiled.

gdetor:
Hi @rLarisch, your article is now online http://rescience.github.io/bibliography/Larisch_2019.html
This issue is now closing.

